PR #20, opened 2021-08-21
Add up- and downvote tooltip

Pixelguymm:
Shows the map's up- and downvotes when hovering over its rating.


Top-Cat:
This seems mostly harmless, I'll get to it when I'm out of a cdn hole

PR #22, opened 2021-08-22
Enforce https redirect

ByteAlex:
At the moment http://beatsaver.com serves an website without TLS, instead it should return a redirect to https.
This PR aims to fix that by enforcing an https redirect via the KTor "HttpsRedirect" feature.

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/server.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/server.kt
@@ -40,6 +40,7 @@ import io.ktor.features.DataConversion
 import io.ktor.features.NotFoundException
 import io.ktor.features.ParameterConversionException
 import io.ktor.features.StatusPages
+import io.ktor.features.HttpsRedirect
 import io.ktor.features.XForwardedHeaderSupport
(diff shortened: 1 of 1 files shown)


ByteAlex:
Though that only happens on browsers which never have visited the https site before from what I can tell. But if you initially did not visit the https site (i.e. by just typing beatsaver.com) it will server a http website. I just noticed that yesterday when I installed another browser to test a browser extension issue.

Top-Cat:
?
```bash
# curl -v http://beatsaver.com

< HTTP/1.1 301 Moved Permanently
< Server: nginx
< Date: Sun, 22 Aug 2021 23:41:52 GMT
< Content-Type: text/html
< Content-Length: 178
< Connection: keep-alive
< Location: https://beatsaver.com
```

Ktor isn't actually doing SSL termination either, this PR would just create a redirect loop for clients.

ByteAlex:
Maybe you changed something in the past few days in your proxy settings, but I'm certainly sure, that I was having a http website served because I have noticed that the buttons for my extension weren't showing up and that was due to a URL check on "https://".
But fair enough - if it serves strictly ssl now, thats fine.

PR #76, opened 2022-02-20
Add hover titles to map sidebar colors

Pixelguymm:


src/jsMain/kotlin/io/beatmaps/index/beatmapInfo.kt
```diff
@@ -67,18 +67,24 @@ class BeatmapInfo : RComponent<BeatmapInfoProps, BeatMapInfoState>() {
...
         props.map?.let { map ->
-            val classes = listOfNotNull(
-                "beatmap",
+            val mapAttributes = listOfNotNull(
                 if (map.ranked) "ranked" else null,
                 if (map.qualified && !map.ranked) "qualified" else null,
                 if (map.curator != null) "curated" else null
+            )
+
+            val classes = listOf(
```
Top-Cat, line 76: Not technically wrong but I feel like the spread operator is a bit overkill when you can just
```kotlin
val classes = mapAttributes.plus("beatmap").joinToString(" ")
```
or if you prefer having beatmap first
```kotlin
val classes = listOf("beatmap").plus(mapAttributes).joinToString(" ")
```

PR #77, opened 2022-02-20
Add hover titles to map sidebar colors

Pixelguymm:


src/jvmMain/sass/main.scss
```diff
@@ -18,7 +18,7 @@ $bs-green: #008055;
 $bs-red: #b52a1c;
 $bs-orange: #bd5500;
 $bs-yellow: #fff301;
-$bs-purple: #6610f2;
+$bs-purple: #454088;
```
Top-Cat, line 21: Someone will definitely point this out

PR #95, opened 2022-03-26
Add song length and BPM to map detail page

XAce1337manX:
https://github.com/beatmaps-io/beatsaver-main/issues/69. Nice.

The details


Top-Cat:
At some point I'll probably rearrange this panel so it doesn't take up so much space but for now this is fine.

helencarnate:
> At some point I'll probably rearrange this panel so it doesn't take up so much space but for now this is fine.

I can also look at this as part of the map detail page design work.

PR #96, opened 2022-03-27
Add bot map flagging, disallow self-curation

Pixelguymm:


src/jsMain/kotlin/io/beatmaps/maps/mapinfo.kt
```diff
@@ -2,11 +2,7 @@ package io.beatmaps.maps
 
 import external.Axios
 import external.generateConfig
-import io.beatmaps.api.CurateMap
-import io.beatmaps.api.MapDetail
-import io.beatmaps.api.MapInfoUpdate
-import io.beatmaps.api.SimpleMapInfoUpdate
-import io.beatmaps.api.StateUpdate
+import io.beatmaps.api.*
```
Top-Cat, line 5: Please tell intelliJ to not

src/jsMain/kotlin/io/beatmaps/maps/mapinfo.kt
```diff
@@ -171,7 +173,7 @@ class MapInfo : RComponent<MapInfoProps, MapInfoState>() {
                                     }
                                 }
 
-                                if (userData?.curator == true) {
+                                if (userData?.curator == true && !props.isOwner) {
```
Top-Cat, line 176: Probably want to add server-side validation for that

Pixelguymm: Good idea.

src/jsMain/kotlin/io/beatmaps/maps/mapinfo.kt
Top-Cat: https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports
(But for Kotlin instead of Java)

PR #107, opened 2022-05-01
Add map/user references + small UI tweak

### LINKED ISSUE #106: Tagging mappers in descriptions
It would be really useful if we could create a hyperlink to a mapper's profile page using `@[username]`. It would remove the need to add a full URL for every mapper who contributed.

Pixelguymm:
Resolves #106


src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
```diff
@@ -175,6 +176,18 @@ fun Route.mapController() {
         }
     }
 
+    get<UserController.RedirectName> {
+        transaction {
+            User.select {
+                User.name eq it.username.lowercase()
```
Top-Cat, line 182: Should be using uniqueName not name and probably only find active accounts.
See https://github.com/beatmaps-io/beatsaver-main/blob/master/src/jvmMain/kotlin/io/beatmaps/api/users.kt#L690
`(User.uniqueName eq it.name) and User.active`

PR #108, opened 2022-05-04
Improve regex

Pixelguymm:
Fixes punctuation causing user/map references to break.


Top-Cat:
I think this looks ok but can you provide some examples this changes/fixes?

Pixelguymm:
Before the change: 
image
And after: 
image

PR #128, opened 2022-07-05
Implement new alert system

Pixelguymm:
Currently, the only visible change will be that the alert messages look slightly different. Reworked alerts GUI will follow.


src/commonMain/resources/db/migration/V24__Alerts.sql
```diff
@@ -0,0 +1,76 @@
...
+) TABLESPACE pg_default;
+
+ALTER TABLE public.alert_recipient OWNER TO beatmaps;
+
+CREATE UNIQUE INDEX alert_link
+    ON public.alert_recipient USING btree
+        ("recipientId" ASC NULLS LAST, "alertId" ASC NULLS LAST)
+    TABLESPACE pg_default;
+
+INSERT INTO public.alert (head, body, "type", "sentAt") SELECT
```
Top-Cat, line 63: This query fails on prod

Fun fact: Not all moderator actions are map deletions

src/jsMain/kotlin/io/beatmaps/index/sharedMapComponents.kt
```diff
@@ -35,10 +35,8 @@ val botInfo = functionComponent<BotInfoProps> { props ->
             +text
         }
 
-    if (score < -4 || props.automapper == true) {
+    if (score < 0 && props.automapper != false) {
```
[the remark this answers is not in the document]
Pixelguymm: You're right, checking the score is unnecessary.

src/commonMain/resources/db/migration/V24__Alerts.sql
```diff
@@ -0,0 +1,78 @@
...
+     'Removal Notice',
+     'Your map #' || modlog."mapId" || ': **' || beatmap.name || '** has been removed by a moderator.' || E'\n' ||
+     'Reason: *"' || substr(modlog.action, 18, length(modlog.action) - 21) || '"*',
```
Top-Cat, line 66: This still doesn't work, don't know if the spacing is different locally for any reason but I'd recommend changing this line to:
```sql
'Reason: *' || (modlog.action::json->'reason')::text || '*',
```

Pixelguymm: Good call, I knew there had to be a better way.

src/jvmMain/kotlin/io/beatmaps/api/alerts.kt
```diff
@@ -0,0 +1,138 @@
...
+    @Location("/markall")
+    data class MarkAll(val api: AlertsApi)
+}
+
+data class MarkAlert(val id: Int, val read: Boolean)
+
+data class MarkAllAlerts(val read: Boolean)
+
+fun Route.alertsRoute() {
+    options<AlertsApi.Unread> {
```
Top-Cat, line 43: These options routes only need to be present on routes intended to be called from other sites (ie rabbit previewer calls /api/maps/{key} to get title / download url). At this time no authenticated routes need this.

They would also need the header adding to get requests which hasn't been done.

src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -450,22 +446,7 @@ fun Route.userRoute() {
...
-            val alerts = transaction {
-                ModLog.join(Beatmap, JoinType.INNER, Beatmap.id, ModLog.opOn).select {
-                    (Beatmap.uploader eq targetId) and
-                        (ModLog.type inList listOf(ModLogOpType.Unpublish, ModLogOpType.Delete).map { it.ordinal })
-                }.orderBy(ModLog.opAt, SortOrder.DESC).limit(30).map { Alert.from(it, cdnPrefix()) }
-            }
-
-            call.respond(alerts)
-        }
+        call.respondRedirect("/api/alerts/unread" + if (it.id != null) "/${it.id}" else "")
```
Top-Cat, line 449: What purpose does this redirect serve? The new response is mostly incompatible so any users calling this endpoint would receive an error anyway

Pixelguymm: I figured we might not want to remove routes, but I'll get rid of it if it's not necessary.

Top-Cat: If we wanted to keep it we should keep the response here compatible but as it's an internal route I'd delete it.

Top-Cat:
Rebased and merged

PR #134, opened 2022-07-13
Implement new alerts GUI

Pixelguymm:
Requires beatmaps-io/beatsaver-common-mp#3.

### DIFF
--- a/src/jsMain/kotlin/io/beatmaps/user/alerts.kt
+++ b/src/jsMain/kotlin/io/beatmaps/user/alerts.kt
@@ -11,6 +11,7 @@ import io.beatmaps.index.coloredCard
 import io.beatmaps.setPageTitle
 import io.beatmaps.util.textToContent
 import kotlinx.html.js.onClickFunction
+import kotlinx.html.title
 import react.RBuilder
 import react.RComponent
 import react.RProps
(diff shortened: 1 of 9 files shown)


src/jsMain/kotlin/io/beatmaps/user/alerts.kt
```diff
@@ -71,26 +85,33 @@ class AlertsPage : RComponent<AlertsPageProps, AlertsPageState>() {
...
+                        span {
+                            b {
+                                +alert.head
+                            }
+                            +" - "
+                            TimeAgo.default {
+                                attrs.date = alert.time.toString()
+                            }
+                        }
+                        div("ms-auto flex-shrink-0") {
```
Top-Cat, line 103: Probably should only show for user as admins can't perform this action

src/jvmMain/kotlin/io/beatmaps/api/alerts.kt
```diff
@@ -99,7 +101,14 @@ fun Route.alertsRoute() {
                 } > 0
             }
 
-            call.respond(if (result) HttpStatusCode.OK else HttpStatusCode.BadRequest)
+            if (result) {
+                call.sessions.set(
+                    user.copy(alerts = user.alerts?.let { if (req.read) it - 1 else it + 1 })
```
Top-Cat, line 106: I would probably just query the new value in case an alert has been added since the last time we checked

PR #141, opened 2022-07-29
Oauth2 implementation with a single client

NSGolova:
/authorize page UI
image

"Client" implementation https://github.com/BeatLeader/beatleader-server/pull/4

Things I'm not sure in:
- Client `secret` replacement. Is there some way to use secure variable storage?
- Usage of `InMemoryTokenStore`. Should tokens be saved to DB?


Top-Cat:
- Client secret replacement. Is there some way to use secure variable storage?
  - Most secrets use environment variables (see database passwords etc) but I'd probably move clients to the db as I suggested in discord (and not use InMemoryClient so that clients can be updated at runtime, it shouldn't be too hard to provide an object that queries a table and returns a client)
- Usage of InMemoryTokenStore. Should tokens be saved to DB?
  - Your use case doesn't require continued token use. Unless you see a reason you'd want to request the user's data again later then given oauth currently ONLY lets you make calls to one endpoint (identity) it's fine to leave the access tokens / refresh tokens in memory. I had a quick look at your code and I can see you're storing them though, which suggests you might care if they stop working :p

Couple other points:
- You have some wildcard imports, please adjust your IntelliJ settings and revert these
- You should probably implement tokenInfoCallback, it's very weird that the identity endpoint is returning the user's id as a "username" and while I understand it might make sense for the oauth code to do this internally we can correct the field name to "id" using the callback (and even provide some more data like the user's current name and their avatar as you seem to have put the effort in to add that to the Identity object)

Now for the potentially real PITA:
My understanding is that normally in the oauth flow there's a step to confirm with the user that they want to allow BeatLeader (or anyone else that uses login with beatsaver) whatever scopes are being requested. If the user is already logged in currently they just get magically redirected back and it's done. Open to hearing what you think but ideally we would have some list stored in the user's session of clients they've approved and then have another screen that appears before oauth continues.

NSGolova:
Implemented tokenInfoCallback with data from the identity object.

Added page for logged-in users and info about data being shared. I've implemented only the `identity` scope with id, name, and avatar. Client_id is used as the client name, not sure if it should be numeral with the client name in a separate field.

Fixed oauth2 with Discord: it will send a hex-encoded redirect in the &state= param. Login is sending nonce there currently, but it is not checked, so I can safely send oauth2 data instead.

Wildcards are fixed. Maybe there is a way to add `idea` settings to the repository?

I used the env variable for the client secret. I'll look into adding it to the DB but it will probably take me the same time as everything else XD

src/jvmMain/kotlin/io/beatmaps/login/discord.kt
```diff
@@ -2,7 +2,6 @@ package io.beatmaps.login
 
 import com.toxicbakery.bcrypt.Bcrypt
 import io.beatmaps.api.alertCount
-import io.beatmaps.common.Config
```
Top-Cat, line 5: Uh...

src/jvmMain/kotlin/io/beatmaps/login/discord.kt
```diff
@@ -20,25 +19,30 @@ import org.jetbrains.exposed.sql.and
...
+            this.parameters["state"] = state
+        }
+    },
 )
 
-class SimpleUserPrincipal(val user: UserDao, val alertCount: Int) : Principal
+class SimpleUserPrincipal(val user: UserDao, val alertCount: Int, val redirect: String) : Principal
 
 fun Application.installDiscordOauth() {
     val baseName = System.getenv("BASE_URL") ?: Config.basename
```
Top-Cat, line 40: Right here ^

NSGolova: Oops, ktlint deleted according to local uncommitted changes XD

PR #143, opened 2022-07-31
Add improved alerts indicator

Pixelguymm:
Known issue: Due to caching (presumably), alert count can display incorrectly when moving between pages. Refreshing fixes this.


Top-Cat:
Were we not going to just stick to the red dot?
You could avoid using nav space and just move the dot out instead of the whole link? Most people still won't have much reason to see alerts.

Pixelguymm:
The bell was part of Helen's mock, though it also had a dropdown menu that turned out to be very impractical to implement. I'd be fine with dropping the idea.

Top-Cat:
Was it a dropdown of actual alerts? I can see how you could do that but I'd probably not prioritise it.

Pixelguymm:
Yes, this is the design we settled on:
Alert Dropdown Mock

PR #146, opened 2022-08-02
Reverted try-catch on state parsing

NSGolova:
- Checked also for Oauth2 param in case of gibberish does parse to something


Top-Cat:
That doesn't throw an error

PR #147, opened 2022-08-03
Update grey uncertainty bar

XAce1337manX:
This is Kansas' grey bar formula.

Gist of changes:
* Changes uncertainty to be based only on vote count.
  * This makes even split votes no longer show 0 grey.
  * This makes it easier to determine how many votes a map has at a glance.
* Grey starts large with additional uncertainty for low vote count and diminishes quicker.
  * This stops 0/1 maps being covered with 60% of yellow and reduces visual impact of initial votes.

Comparison:
image

### DIFF
--- a/src/jsMain/kotlin/io/beatmaps/index/beatmapInfo.kt
+++ b/src/jsMain/kotlin/io/beatmaps/index/beatmapInfo.kt
@@ -24,9 +24,8 @@ import react.dom.p
 import react.dom.small
 import react.dom.span
 import react.setState
-import kotlin.math.abs
 import kotlin.math.floor
-import kotlin.math.log10
+import kotlin.math.log
 import kotlin.math.pow
 
 external interface BeatmapInfoProps : RProps {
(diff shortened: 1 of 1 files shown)


src/jsMain/kotlin/io/beatmaps/index/beatmapInfo.kt
```diff
@@ -94,8 +93,12 @@ class BeatmapInfo : RComponent<BeatmapInfoProps, BeatMapInfoState>() {
 
                     div {
                         val totalVotes = (map.stats.upvotes + map.stats.downvotes).toDouble()
-                        val rawScore = map.stats.upvotes / totalVotes
-                        val uncertainty = abs((rawScore - 0.5) * 2.0.pow(-log10(totalVotes + 1)))
+                        var uncertainty = 2.0.pow(-log(totalVotes / 2 + 1, 3.0))
+                        val weightedRange = 25
+                        val weighting = 2
+                        if ((totalVotes + weighting) < weightedRange) {
+                            uncertainty += (1 - uncertainty) * (1 - (totalVotes + weighting) * (1 / weightedRange))
```
Top-Cat, line 100: `(1 / weightedRange)` will use integer division and always be 0, making uncertainty always 1. You then get a lovely division by 0 below. All of which means any map under 23 votes has no grey bar at all

XAce1337manX: A classic textbook mistake. 🤦

PR #148, opened 2022-08-03
Add improved alerts indicator

Top-Cat:
Known issue: Due to caching (presumably), alert count can display incorrectly when moving between pages. Refreshing fixes this.


Top-Cat:
You saw nothing

PR #152, opened 2022-08-21
Add Playlist feed

Pixelguymm:
This was made with very little testing data, so features that require a lot of data (like infinite scrolling for example) might need to be tweaked.


src/jvmMain/kotlin/io/beatmaps/api/playlist.kt
```diff
@@ -213,36 +226,40 @@ fun Route.playlistRoute() {
         val searchFields = PgConcat(" ", Playlist.name, Playlist.description)
         val searchInfo = parseSearchQuery(it.q, searchFields)
         val actualSortOrder = searchInfo.validateSearchOrder(it.sortOrder)
+        val sortArgs = when (actualSortOrder) {
+            SearchOrder.Relevance -> listOf(searchInfo.similarRank to SortOrder.DESC, Playlist.avgScore to SortOrder.DESC, Playlist.createdAt to SortOrder.DESC)
+            SearchOrder.Rating -> listOf(Playlist.avgScore to SortOrder.DESC, Playlist.createdAt to SortOrder.DESC)
```
Top-Cat, line 231: Uh yeah... so. Sorting by average score of a playlist means the database needs to know the average score for every playlist. To do that it obviously has to cross join every playlist with every map and aggregate the whole database potentially many times over depending on how many playlists we have.

There are some other issues with the query that exposed is generating (you have to be far more careful to get an efficient query with all these joins) but while those may be fixable this fundamentally isn't. It's why the user table has an upvote column (it's used for sorting the mappers list).

Probably best to include any data we want to sort or filter by into the playlists table. My current list is average score, min nps, max nps, verified and emptyness (could just store the count itself). Honestly at that point there's a fairly good argument for including the rest of the stats too as we're already going to have to have a system to update these.

Random aside: I'm fairly sure the from/to filters are bugged and were meant to filter when the playlist was made not the maps in them.

Top-Cat:
imo would be nicer if these had set widths so they looked more like a table
image

Pixelguymm:
> imo would be nicer if these had set widths so they looked more like a table

Agreed, I didn't think the width would vary this much.

Top-Cat:
1. Infinite scroll isn't working. Not sure why you've switched from `rangeTo` to `until` as this excludes the end of the range.
```
> 0.rangeTo(1)
0..1
> 0.until(1)
0..0
```
2. Linter appears to be upset at these imports:
```
src\jsMain\kotlin\io\beatmaps\index\beatmapInfo.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end
src\jsMain\kotlin\io\beatmaps\index\main.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end
src\jsMain\kotlin\io\beatmaps\maps\mapinfo.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end
src\jsMain\kotlin\io\beatmaps\modlog\main.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end
src\jsMain\kotlin\io\beatmaps\user\alerts.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end
src\jsMain\kotlin\io\beatmaps\playlist\main.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end
src\jsMain\kotlin\io\beatmaps\shared\sharedComponents.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end
```

Pixelguymm:
1. I'm pretty sure I did that to fix a bug where the loading placeholders would stay on the page when there were no results, even after it was done loading. I'm not sure why that fixed it though, I'll try to find another way.
2. Oops.

src/jsMain/kotlin/io/beatmaps/playlist/playlistTable.kt
```diff
@@ -0,0 +1,193 @@
...
+        loadNextPage()
+    }
+
+    private fun lastPage() = max(state.visiblePages.last, state.pages.maxByOrNull { it.key }?.key ?: 0)
+
+    override fun RBuilder.render() {
+        if (props.visible == false) return
+
+        if (props.own == true) {
+            routeLink("/playlists/new", className = "btn btn-success btn-sm") {
```
Top-Cat, line 231: I don't really have an answer but this button looks lost now. Either needs to be bigger, have some dialog around it or.... something

Image for people so they don't have to get this running to see:
image

src/commonMain/kotlin/io/beatmaps/api/search.kt
```diff
@@ -1,8 +1,19 @@
 // ktlint-disable filename
```
Top-Cat, line 1: Nitpick: not needed anymore

src/jvmMain/kotlin/io/beatmaps/api/playlist.kt
```diff
@@ -176,23 +190,33 @@ fun Route.playlistRoute() {
 
         val playlists = transaction {
             Playlist
+                .joinMaps {
+                    Beatmap.deletedAt.isNull()
+                }
                 .joinPlaylistCurator()
                 .joinOwner()
+                .slice(Playlist.columns + User.columns + playlistStats)
```
Top-Cat, line 198: + curatorAlias.columns

src/jvmMain/kotlin/io/beatmaps/api/playlist.kt
```diff
@@ -214,34 +238,49 @@ fun Route.playlistRoute() {
...
         val actualSortOrder = searchInfo.validateSearchOrder(it.sortOrder)
+        val sortArgs = when (actualSortOrder) {
+            SearchOrder.Relevance -> listOf(searchInfo.similarRank to SortOrder.DESC, Playlist.createdAt to SortOrder.DESC)
+            SearchOrder.Rating, SearchOrder.Latest -> listOf(Playlist.createdAt to SortOrder.DESC)
+            SearchOrder.Curated -> listOf(Playlist.curatedAt to SortOrder.DESC_NULLS_LAST, Playlist.createdAt to SortOrder.DESC)
+        }.toTypedArray()
 
         newSuspendedTransaction {
             val playlists = Playlist
+                .joinMaps(type = JoinType.LEFT) {
```
Top-Cat, line 249: This is the default and `Beatmap.deletedAt.isNull()` should possibly be the default for state

PR #153, opened 2022-08-28
Add modlog button to profile page

Pixelguymm:
As requested by Blookcloak


src/jsMain/kotlin/io/beatmaps/user/main.kt
```diff
@@ -165,6 +165,12 @@ class ProfilePage : RComponent<ProfilePageProps, ProfilePageState>() {
                                 i("fas fa-list") { }
                                 +"Playlist"
                             }
+                            if (props.userData?.admin == true) {
+                                a("/modlog?user=${state.userDetail?.name}", classes = "btn btn-secondary me-2") {
```
Top-Cat, line 169: routeLink?

PR #154, opened 2022-08-29
Small tweaks and fixes

Pixelguymm:


src/jvmMain/kotlin/io/beatmaps/util/playliststats.kt
```diff
@@ -49,8 +49,8 @@ fun Application.playlistStats() {
                     Playlist.id eq playlistId
                 }) {
                     it[totalMaps] = totalMapsVal
-                    it[minNps] = minNpsVal
-                    it[maxNps] = maxNpsVal
+                    it[minNps] = minNpsVal.setScale(2)
```
Top-Cat, line 52: Would probably recommend rounding these on the front end.
Both in case we change our minds and that this only affects new changes.

PR #156, opened 2022-08-30
Fix Curated sorting option

Pixelguymm:


src/jsMain/kotlin/io/beatmaps/shared/search.kt
```diff
@@ -269,12 +271,13 @@ open class Search<T>(props: SearchProps<T>) : RComponent<SearchProps<T>, SearchS
...
                         attrs.onChangeFunction = {
                             setState {
-                                order = SearchOrder.fromInt(sortRef.current?.selectedIndex ?: 0) ?: SearchOrder.Relevance
+                                order = SearchOrder.fromString(sortRef.current?.let { it.options[it.selectedIndex]?.id } ?: "") ?: SearchOrder.Relevance
                             }
                         }
                         SearchOrder.values().filter { props.sortOrderTarget in it.targets }.forEach {
                             option {
                                 attrs.selected = state.order == it
+                                attrs.id = it.toString()
```
Top-Cat, line 280: I think one would traditionally use the value attribute (ignoring the fact this is already the innertext/html of the option anyway)

Issue #8, opened 2021-08-05
Allow hash lookup on unpublished hashes


Top-Cat:
Return current map data when looking up by hash as long as there is a published version and the map is not deleted even if the hash is for an unpublished version

Top-Cat:
Done

PR #157, opened 2022-09-07
Add basic follow system

Pixelguymm:
Requires beatmaps-io/beatsaver-common-mp#8

### DIFF
--- a/src/commonMain/resources/db/migration/V30__Follows.sql
+++ b/src/commonMain/resources/db/migration/V30__Follows.sql
@@ -1,0 +1,27 @@
+CREATE SEQUENCE public."follows_followId_seq"
+    INCREMENT 1
+    START 1
+    MINVALUE 1
+    NO MAXVALUE
+    CACHE 1;
+
+ALTER TABLE public."follows_followId_seq" OWNER TO beatmaps;
+
+CREATE TABLE public.follows
+(
+    "followId" integer NOT NULL DEFAULT nextval('"follows_followId_seq"'::regclass),
+    "userId" integer,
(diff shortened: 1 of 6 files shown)


src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -639,7 +667,16 @@ fun Route.userRoute() {
             (User.id eq it.id) and User.active
         }
 
-        val userDetail = UserDetail.from(user, stats = statsForUser(user)).let {
+        val following = call.sessions.get<Session>()?.let { sess ->
+            if (it.id == sess.userId) null
+            else transaction {
+                Follows.select {
```
Top-Cat, line 673: Would make sense to me to combine this and the above query (at least into one transaction, but one query also seems trivial)

src/commonMain/resources/db/migration/V30__Follows.sql
```diff
@@ -0,0 +1,27 @@
+CREATE SEQUENCE public."follows_followId_seq"
+    INCREMENT 1
+    START 1
+    MINVALUE 1
+    NO MAXVALUE
+    CACHE 1;
+
+ALTER TABLE public."follows_followId_seq" OWNER TO beatmaps;
+
+CREATE TABLE public.follows
```
Top-Cat, line 10: Need a unique index

Top-Cat: (Both to stop people following users twice and because right now checking if you follow a user requires scanning this whole table)

Issue #9, opened 2021-08-05
[Request] Add a way to "subscribe" to a mapper


Tootieee:
It'd be nice to have a subscription area similar to youtube in which you subscribe to a mapper, possibly receive notifications for new uploads, and have a separate area where you could see new maps from people you're subscribed to. There doesn't need to be a subscriber count for a mapper, so there's no competitive reason for it. Just something convenient to be able to easily download new maps from your favorite mappers in a single page. Similar to a mod called BeatSync I believe but in browser

helencarnate:
Completed on September 8, 2022 with Pull #157 (Followed Mappers)

PR #158, opened 2022-09-10
Alerts page bug fixes

Pixelguymm:
Requires beatmaps-io/beatsaver-common-mp#9

### DIFF
--- a/src/jsMain/kotlin/io/beatmaps/user/alerts/alert.kt
+++ b/src/jsMain/kotlin/io/beatmaps/user/alerts/alert.kt
@@ -37,6 +37,7 @@ external interface AlertProps : RProps {
 external interface AlertState : RState {
     var height: String?
     var opacity: String?
+    var margin: String?
 }
 
 fun updateAlertDisplay(stats: UserAlertStats) {
@@ -68,6 +69,7 @@ class AlertElement : RComponent<AlertProps, AlertState>() {
         if (nextProps.alert == null) {
             setState {
                 height = "auto"
+                margin = "5px"
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/jsMain/kotlin/io/beatmaps/user/alerts/main.kt
```diff
@@ -54,6 +54,7 @@ class AlertsPage : RComponent<AlertsPageProps, AlertsPageState>() {
 
             setState {
                 alertStats = data
+                hiddenAlerts = listOf()
```
Top-Cat, line 57: Literally can't tell the difference, there shouldn't be one

src/jsMain/kotlin/io/beatmaps/user/alerts/alert.kt
```diff
@@ -78,11 +80,13 @@ class AlertElement : RComponent<AlertProps, AlertState>() {
                 if (props.hidden == true) {
                     height = "0px"
                     opacity = "0"
+                    margin = "-1px 5px" // -1 pixel to account for the border
```
Top-Cat, line 83: If you're trying to account for the border.... why not change the border?

src/jsMain/kotlin/io/beatmaps/user/alerts/main.kt
Pixelguymm: When logging out, logging in again and going to your alerts page, the alerts don't disappear when you click the eye. Once you change the a filter or go to your read alerts, they do.

src/jsMain/kotlin/io/beatmaps/user/alerts/alert.kt
Pixelguymm: As far as I could tell, the border width couldn't be animated with `transition`.

Top-Cat: `transition: opacity .2s, height .2s linear .2s, border-width .2s linear .2s`
works just fine if you actually define that property

src/jsMain/kotlin/io/beatmaps/user/alerts/main.kt
Top-Cat: Ah, would probably prefer this be handled on line 192
`hiddenAlerts = (hiddenAlerts ?: emptyListOf()).plus(id)`

Issue #10, opened 2021-08-06
Improve mapper search


Top-Cat:
- [x] Include songAuthorName in search index
- [x] (Optional) Add some method of restricting search to single mapper `mapper:reaxt`

Top-Cat:
Also possibly add a dedicated api endpoint for finding mapper info

Top-Cat:
I forgot that I completed this

PR #159, opened 2022-09-16
Update profile pages

Pixelguymm:
Requires beatmaps-io/beatsaver-common-mp#10.

### DIFF
--- a/src/commonMain/kotlin/io/beatmaps/api/users.kt
+++ b/src/commonMain/kotlin/io/beatmaps/api/users.kt
@@ -20,7 +20,7 @@ data class UserDetail(
     val avatar: String,
     val stats: UserStats? = null,
     val followData: UserFollowData? = null,
-    val type: AccountType = AccountType.SIMPLE,
+    val type: AccountType,
     val email: String? = null,
     val uploadLimit: Int? = null,
     val admin: Boolean? = null,
--- a/src/commonMain/kotlin/io/beatmaps/api/users.kt
+++ b/src/commonMain/kotlin/io/beatmaps/api/users.kt
@@ -13,18 +13,30 @@ enum class AccountType {
 data class UserDetail(
     val id: Int,
     val name: String,
+    val description: String,
     val uniqueSet: Boolean = true,
     val hash: String? = null,
     val testplay: Boolean? = null,
     val avatar: String,
     val stats: UserStats? = null,
+    val followData: UserFollowData? = null,
     val type: AccountType,
     val email: String? = null,
     val uploadLimit: Int? = null,
+    val admin: Boolean? = null,
     val curator: Boolean? = null,
     val verifiedMapper: Boolean = false,
-    val following: Boolean? = null
 ) { companion object }
+@Serializable
+data class FollowerData(
+    val id: Int,
+    val name: String,
+    val avatar: String,
+    val maps: Int,
+    val admin: Boolean? = null,
+    val curator: Boolean? = null,
+    val verifiedMapper: Boolean = false,
+)
 @Serializable
 data class UserStats(
     val totalUpvotes: Int,
@@ -39,13 +51,19 @@ data class UserStats(
     val diffStats: UserDiffStats? = null
 )
 @Serializable
+data class UserFollowData(
+    val followers: Int,
+    val follows: Int?,
+    val following: Boolean?
+)
+@Serializable
 data class UserDiffStats(val total: Int, val easy: Int, val normal: Int, val hard: Int, val expert: Int, val expertPlus: Int)
 @Serializable
 data class BeatsaverLink(val linked: Boolean) { companion object }
 @Serializable
 data class BeatsaverLinkReq(val user: String, val password: String, val useOldName: Boolean = true)
 @Serializable
-data class UsernameReq(val username: String)
+data class AccountDetailReq(val textContent: String)
 @Serializable
 data class RegisterRequest(val captcha: String, val username: String, val email: String, val password: String, val password2: String)
 @Serializable
--- a/src/commonMain/kotlin/io/beatmaps/api/users.kt
+++ b/src/commonMain/kotlin/io/beatmaps/api/users.kt
@@ -20,7 +20,7 @@ data class UserDetail(
     val avatar: String,
     val stats: UserStats? = null,
     val followData: UserFollowData? = null,
-    val type: AccountType,
+    val type: AccountType = AccountType.SIMPLE,
     val email: String? = null,
     val uploadLimit: Int? = null,
     val admin: Boolean? = null,
@@ -28,26 +28,16 @@ data class UserDetail(
     val verifiedMapper: Boolean = false,
 ) { companion object }
 @Serializable
-data class FollowerData(
-    val id: Int,
-    val name: String,
-    val avatar: String,
-    val maps: Int,
-    val admin: Boolean? = null,
-    val curator: Boolean? = null,
-    val verifiedMapper: Boolean = false,
-)
-@Serializable
 data class UserStats(
-    val totalUpvotes: Int,
-    val totalDownvotes: Int,
-    val totalMaps: Int,
-    val rankedMaps: Int,
-    val avgBpm: Float,
-    val avgScore: Float,
-    val avgDuration: Float,
-    val firstUpload: Instant?,
-    val lastUpload: Instant?,
+    val totalUpvotes: Int = 0,
+    val totalDownvotes: Int = 0,
+    val totalMaps: Int = 0,
+    val rankedMaps: Int = 0,
+    val avgBpm: Float = 0f,
+    val avgScore: Float = 0f,
+    val avgDuration: Float = 0f,
+    val firstUpload: Instant? = null,
+    val lastUpload: Instant? = null,
     val diffStats: UserDiffStats? = null
 )
 @Serializable
--- a/src/jsMain/kotlin/io/beatmaps/shared/sharedComponents.kt
+++ b/src/jsMain/kotlin/io/beatmaps/shared/sharedComponents.kt
@@ -344,7 +344,7 @@ var userCard = functionComponent<UserCardProps> {
         div("d-inline") {
             routeLink("/profile/${it.id}") {
                 h4("mb-1") {
-                    +(it.username)
+                    +it.username
                 }
             }
             p("text-muted mb-1") {
--- a/src/jsMain/kotlin/io/beatmaps/shared/sharedComponents.kt
+++ b/src/jsMain/kotlin/io/beatmaps/shared/sharedComponents.kt
@@ -27,7 +27,9 @@ import react.RReadableRef
 import react.dom.RDOMBuilder
 import react.dom.a
 import react.dom.div
+import react.dom.h4
 import react.dom.i
+import react.dom.img
 import react.dom.input
 import react.dom.jsStyle
 import react.dom.label
@@ -322,6 +324,32 @@ fun RDOMBuilder<DIV>.slider(text: String, currentMin: Float, currentMax: Float, 
         p("m-0 float-end") {
             val maxStr = if (currentMax >= max) "∞" else currentMax.toString()
             +"$currentMin - $maxStr"
+        }
+    }
+}
+
+external interface UserCardProps : RProps {
+    var id: Int
+    var avatar: String
+    var username: String
+    var titles: List<String>
+}
+
+var userCard = functionComponent<UserCardProps> {
+    div("d-flex align-items-center my-2") {
+        img("Profile Image", it.avatar, classes = "rounded-circle me-3") {
+            attrs.width = "50"
+            attrs.height = "50"
+        }
+        div("d-inline") {
+            routeLink("/profile/${it.id}") {
+                h4("mb-1") {
+                    +(it.username)
+                }
+            }
+            p("text-muted mb-1") {
+                +it.titles.joinToString(", ")
+            }
         }
     }
 }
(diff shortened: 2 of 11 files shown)


src/jsMain/kotlin/io/beatmaps/user/account.kt
```diff
@@ -52,13 +55,16 @@ external interface AccountComponentState : RState {
     var errors: List<String>
     var usernameErrors: List<String>
     var username: String
+    var descriptionErrors: List<String>
```
Top-Cat, line 58: I don't understand why this is its own property. It seems to be mutually exclusive with usernameErrors

src/jsMain/kotlin/io/beatmaps/shared/sharedComponents.kt
```diff
@@ -325,3 +327,29 @@ fun RDOMBuilder<DIV>.slider(text: String, currentMin: Float, currentMax: Float,
...
+var userCard = functionComponent<UserCardProps> {
+    div("d-flex align-items-center my-2") {
+        img("Profile Image", it.avatar, classes = "rounded-circle me-3") {
+            attrs.width = "50"
+            attrs.height = "50"
+        }
+        div("d-inline") {
+            routeLink("/profile/${it.id}") {
+                h4("mb-1") {
+                    +(it.username)
```
Top-Cat, line 347: Unnecessary brackets?

src/jsMain/kotlin/io/beatmaps/user/account.kt
```diff
@@ -99,12 +107,12 @@ class AccountComponent : RComponent<AccountComponentProps, AccountComponentState
...
             div("invalid-feedback") {
-                val error = state.usernameErrors.firstOrNull()
-                if (error != null) {
+                val errors = state.usernameErrors + state.descriptionErrors
+                if (errors.isNotEmpty()) {
                     attrs.jsStyle {
                         display = "block"
                     }
-                    +error
+                    +errors.joinToString(" ")
```
Top-Cat, line 115: So, as far as I know this change doesn't do anything because I think the sum total of the two lists can still only ever be one error.
I think the correct behaviour if there were multiple would be to have them on separate lines (see upload/main.kt where we a div is created for each).

src/jsMain/kotlin/io/beatmaps/user/account.kt
```diff
@@ -163,6 +173,65 @@ class AccountComponent : RComponent<AccountComponentProps, AccountComponentState
...
+                                    userLoading = true
+                                }
+
+                                Axios.post<ActionResponse>(
+                                    "${Config.apibase}/users/description",
+                                    AccountDetailReq(state.description),
+                                    generateConfig<AccountDetailReq, ActionResponse>()
+                                ).then {
+                                    if (it.data.success) {
+                                        window.location.reload()
```
Top-Cat, line 214: Would probably be nicer to either call loadState in the parent (pass down a method as a property to trigger this) or allow this to directly modify the user state shown on the page.

src/jsMain/kotlin/io/beatmaps/user/account.kt
Top-Cat: Or actually as there are two different buttons and requests maybe actually don't have them be mutually exclusive?

src/jsMain/kotlin/io/beatmaps/user/account.kt
```diff
@@ -163,6 +173,65 @@ class AccountComponent : RComponent<AccountComponentProps, AccountComponentState
...
+                    }
+                    ref = descriptionRef
+                }
+                div("d-grid") {
+                    button(classes = "btn btn-success", type = ButtonType.submit) {
+                        attrs.onClickFunction = { ev ->
+                            ev.preventDefault()
+
+                            if (props.userDetail.description == state.description) {
+                                setState {
```
Top-Cat, line 199: I'd probably just ignore this tbh, the user wants this as their description and it is.

src/jsMain/kotlin/io/beatmaps/user/main.kt
```diff
@@ -135,137 +148,226 @@ class ProfilePage : RComponent<ProfilePageProps, ProfilePageState>() {
...
+                    div("card-body") {
+                        span {
+                            textToContent((state.userDetail?.description?.take(500) ?: ""))
                         }
-                        img("100 Maps", "https://cdn.discordapp.com/avatars/98334361564246016/01ade7513a63215bb7937d217b766da3.png", classes = "rounded-circle mx-2") {
-                            attrs.width = "75"
-                            attrs.height = "75"
+                        state.userDetail?.stats?.let {
+                            if (it.totalMaps != 0) {
+                                table("table table-dark") {
```
Top-Cat, line 277: Not a huge fan of the table. Mostly the bottom border, looks slightly better with table-borderless. Can possibly roll a custom style.

Also someone's going to complain about losing avg bpm/duration even if they're not that useful.

src/commonMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -13,19 +13,31 @@ enum class AccountType {
...
     val type: AccountType,
     val email: String? = null,
     val uploadLimit: Int? = null,
+    val admin: Boolean? = null,
     val curator: Boolean? = null,
     val verifiedMapper: Boolean = false,
-    val following: Boolean? = null
 ) { companion object }
 @Serializable
+data class FollowerData(
```
Top-Cat, line 31: Worth having a common type these both inherit from as this is all stuff in UserDetail except `maps`

src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -128,23 +133,43 @@ class UsersApi {
...
+    data class Follow(val api: UsersApi)
+
+    @Location("/following/{user}/{page}")
+    data class Following(val user: Int, val page: Long = 0, val api: UsersApi)
+
+    @Location("/followedBy/{user}/{page}")
+    data class FollowedBy(val user: Int, val page: Long = 0, val api: UsersApi)
 }
 
+fun followData(uploaderId: Int, userId: Int?) = UserFollowData(
```
Top-Cat, line 145: Ok, ok, there's nothing wrong with this (except I don't think the way the ifs are written is that readable) but hear me out... What if...

```kotlin
fun followData(uploaderId: Int, userId: Int?): UserFollowData {
    val userFilter = Follows.userId eq uploaderId
    val followerFilter = Follows.followerId eq userId

    val userColumn = countWithFilter(userFilter)
    val followerColumn = if (userId != uploaderId) {
        Op.nullOp()
    } else {
        countWithFilter(followerFilter)
    }

    val bothColumn = if (userId == null || userId == uploaderId) {
        Op.nullOp()
    } else {
        countWithFilter(userFilter and followerFilter).greater(intLiteral(0))
    }

    return Follows
        .slice(userColumn, followerColumn, bothColumn).select {
            (Follows.userId eq uploaderId) or (Follows.followerId eq userId)
        }.single().let {
            UserFollowData(it[userColumn], it[followerColumn], it[bothColumn])
        }
}
```

src/jsMain/kotlin/io/beatmaps/user/main.kt
```diff
@@ -135,137 +148,226 @@ class ProfilePage : RComponent<ProfilePageProps, ProfilePageState>() {
     private fun setFollowStatus(followed: Boolean) {
         Axios.post<UserFollowRequest>("${Config.apibase}/users/follow", UserFollowRequest(state.userDetail?.id ?: 0, followed), generateConfig<UserFollowRequest, String>()).then({
             setState {
-                following = followed
+                followData = UserFollowData(
+                    (followData?.followers ?: 0).let {
+                        if (followed) it + 1 else it - 1
+                    },
+                    followData?.follows ?: 0,
```
Top-Cat, line 155: Probably shouldn't have the default value here. Causes the info line to appear when following/unfollowing users.

src/jvmMain/sass/main.scss
```diff
@@ -873,14 +883,42 @@ div.mapstats {
...
-.user-badges a i {
-  margin-right: 5px;
+.user-info {
+  height: 100%;
+
+  .card-body {
+    display: flex;
+    flex-direction: column;
+
+    > :last-child:not(:first-child) {
```
Top-Cat, line 893: More readable as `:last-child:not(:only-child)`

src/jsMain/kotlin/io/beatmaps/user/main.kt
Pixelguymm: I'll try to fit average bpm and duration in there, though no one has complained so far, and I don't know anyone who uses it.

src/jsMain/kotlin/io/beatmaps/user/main.kt
Pixelguymm: You're right, that wasn't meant to be there.

src/commonMain/kotlin/io/beatmaps/api/users.kt
Pixelguymm: There's no practical use for that right now. I think doing this would only require more code, since the constructors of the derived classes would still need to have every property listed.

Top-Cat: I actually only mentioned this because I saw a practical use (the duplicated code in userTitles)

Pixelguymm: Wouldn't that still need separate functions because of `maps` vs `stats.totalMaps`?

Top-Cat: Not necessarily, if you had a function to pull that property out it wouldn't change how they serialise but that method would work.

Top-Cat: You are right though data class inheritance is a bit painful.

src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -783,10 +783,11 @@ fun Route.userRoute() {
             }
             .groupBy(Follows.id, User.id)
             .map { row ->
-                FollowerData(
-                    row[User.id].value, row[User.uniqueName] ?: row[User.name],
-                    row[User.avatar] ?: "https://www.gravatar.com/avatar/${row[User.hash]}?d=retro",
-                    row[Beatmap.id.count()].toInt(), row[User.admin], row[User.curator], row[User.verifiedMapper]
+                UserDetail(
```
Top-Cat, line 786: UserDetail.from(UserDao.wrap(row), stats = UserStats(totalMaps = row[Beatmap.id.count()].toInt()))
no?

src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -698,6 +761,51 @@ fun Route.userRoute() {
 
         call.respond(UserDetail.from(user, stats = statsForUser(user)))
     }
+
+    fun getFollowerData(page: Long, joinOn: Column<EntityID<Int>>, condition: SqlExpressionBuilder.() -> Op<Boolean>) = transaction {
```
Top-Cat, line 765: Can you not remove a whole join here?

```kotlin
    fun getFollowerData(page: Long, joinOn: Column<EntityID<Int>>, condition: SqlExpressionBuilder.() -> Op<Boolean>) = transaction {
        User
            .join(Beatmap, JoinType.LEFT, User.id, Beatmap.uploader) {
                Beatmap.deletedAt.isNull()
            }
            .slice(
                User.id, User.name, User.uniqueName, User.description, User.avatar, User.discordId, User.hash,
                Beatmap.id.count(), User.admin, User.curator, User.verifiedMapper
            )
            .select {
                User.id.inSubQuery(
                    Follows
                        .slice(joinOn)
                        .select(condition)
                        .limit(20, page)
                        .orderBy(Follows.since, SortOrder.DESC)
                )
            }
            .groupBy(User.id)
            .map { row ->
                UserDetail.from(UserDao.wrapRow(row), stats = UserStats(totalMaps = row[Beatmap.id.count()].toInt()))
            }
    }
```

Top-Cat: I was sort of wrong here, we need the join so we can resort on since at the outer level

Top-Cat: More code incoming. This does perform consistently slightly better than the existing code. Uses a subquery in select which allows us to get the since column out rather than the in above which doesn't.

```kotlin
    fun getFollowerData(page: Long, joinOn: Column<EntityID<Int>>, condition: SqlExpressionBuilder.() -> Op<Boolean>) = transaction {
        val followsSubquery = Follows
            .slice(joinOn, Follows.since)
            .select(condition)
            .limit(page)
            .orderBy(Follows.since, SortOrder.DESC)
            .alias("fs")

        followsSubquery
            .join(User, JoinType.LEFT, followsSubquery[joinOn], User.id)
            .join(Beatmap, JoinType.LEFT, User.id, Beatmap.uploader) {
                Beatmap.deletedAt.isNull()
            }
            .slice(
                User.id, User.name, User.uniqueName, User.description, User.avatar, User.discordId, User.hash,
                Beatmap.id.count(), User.admin, User.curator, User.verifiedMapper
            )
            .selectAll()
            .groupBy(User.id, followsSubquery[Follows.since])
            .orderBy(followsSubquery[Follows.since], SortOrder.DESC)
            .map { row ->
                UserDetail.from(UserDao.wrapRow(row), stats = UserStats(totalMaps = row[Beatmap.id.count()].toInt()))
            }
    }
```

Top-Cat:
Infinite scroll on followers also doesn't work. I can look at fixing this if you want.
The initial issue at least seems to be because it's hooking the scroll event of the window whereas you're scrolling within the modal.  So that needs adding as a property to the infinite scroll class to point it at the scroll parent.

Pixelguymm:
> I can look at fixing this if you want.

Sure, then I'll update the query.

PR #161, opened 2022-09-26
Add GitHub to social link parser

Caeden117:
clearly GitHub is a very important social platform that demands beatsaver bio support. This PR (attempts to) map `gh@<username>` to `https://github.com/<username>`

I will be completely transparent: This was done with the github online text editor lmao


Top-Cat:
That face when
image

Caeden117:
From the looks of things, it would be a problem with Twitch and Twitter, assuming those platforms allow `-` and related characters.

The YouTube regex seems to not have this problem, so I'll probably copy from that

Issue #11, opened 2021-08-06
Restrict uploader names to alphanumeric chars or add a new field


n3tman:
Previous username rules by Jellyfish (from `#modding-announcements`)
> Usernames may now only include alphanumeric characters or a `-` character.

Any string in a username doesn't allow to create URL-friendly post tags on BSaber.

Example uploader: https://beatsaver.com/profile/4284389
Current `uploader.name` from the API: `AndroGeek | Beat Saber Mapper`
Previously it was just `androgeek`: https://bsaber.com/tag/androgeek/

A possible solution would be to ask users who logged in through Discord to enter a separate name that follows old naming rules.

Top-Cat:
Users are now being asked to pick a name that matches `^[\._\-A-Za-z0-9]{3,}$`
Anyone who had linked an old beatsaver account has it pre-filled but it's going to take some time for discord-only accounts to get filled in

PR #166, opened 2022-10-22
Add suspension system

Pixelguymm:
Requires beatmaps-io/beatsaver-common-mp#11.


src/jvmMain/kotlin/io/beatmaps/login/main.kt
```diff
@@ -73,7 +73,8 @@ data class Session(
     val canLink: Boolean = false,
     val alerts: Int? = null,
     val curator: Boolean = false,
-    val oauth2ClientId: String? = null
+    val oauth2ClientId: String? = null,
+    val suspended: Boolean = false
 ) {
     constructor(userId: Int, hash: String?, userEmail: String, userName: String, testplay: Boolean, steamId: Long?, oculusId: Long?, admin: Boolean, uniqueName: String?, canLink: Boolean, alerts: Int?, curator: Boolean) :
```
Top-Cat, line 79: Need to add a constructor here to upgrade old sessions

src/jvmMain/kotlin/io/beatmaps/login/main.kt
```diff
@@ -76,8 +76,10 @@ data class Session(
     val oauth2ClientId: String? = null,
     val suspended: Boolean = false
 ) {
+    constructor(userId: Int, hash: String?, userEmail: String, userName: String, testplay: Boolean, steamId: Long?, oculusId: Long?, admin: Boolean, uniqueName: String?, canLink: Boolean, alerts: Int?, curator: Boolean, suspended: Boolean) :
```
Top-Cat, line 79: Nope, you need to upgrade old sessions. They won't have your new field. They will have all the old fields.

PR #168, opened 2022-11-03
NPS always displays 2 decimals

XAce1337manX:
symphonic suggestion
a


Top-Cat:
Should use the cross-platform helper `fixedStr` here: https://github.com/beatmaps-io/beatsaver-common-mp/blob/master/src/commonMain/kotlin/io/beatmaps/common/math.kt

PR #176, opened 2022-12-27
Bookmarks

Pixelguymm:
Requires beatmaps-io/beatsaver-common-mp#12.


src/commonMain/resources/db/migration/V40__Bookmarks.sql
```diff
@@ -0,0 +1,25 @@
...
+UPDATE public.playlist
+SET type = 'Public'
+WHERE public;
+
+UPDATE public.playlist
+SET type = 'Private'
+WHERE NOT public;
+
+ALTER TABLE public.playlist
+DROP COLUMN public;
```
Top-Cat, line 22: In an ideal world we'd probably leave this column initially in case there was some problem that meant we wanted to roll back. Big points for the no-op update above too setting the type to the default value again 😛

src/jsMain/kotlin/io/beatmaps/playlist/edit.kt
```diff
@@ -132,7 +133,7 @@ class EditPlaylist : RComponent<PlaylistEditProps, PlaylistEditState>() {
                             fun sendForm(data: FormData) {
                                 data.append("name", nameRef.current?.value ?: "")
                                 data.append("description", descriptionRef.current?.value ?: "")
-                                data.append("public", (publicRef.current?.checked ?: false).toString())
+                                data.append("type", if (publicRef.current?.checked == true) "Public" else "Private")
```
Top-Cat, line 136: Probably shouldn't hardcode strings here that relate to enum values

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
```diff
@@ -0,0 +1,157 @@
...
+    val bookmarksId = getBookmarksId(userId)
+
+    return PlaylistMap
+        .select { (PlaylistMap.mapId eq mapId) and (PlaylistMap.playlistId eq bookmarksId) }
+        .count() > 0
+}
+
+fun Route.bookmarkRoute() {
+    options<BookmarksApi.Add> {
+        call.response.header("Access-Control-Allow-Origin", "*")
```
Top-Cat, line 74: What's the rationale for having cors headers for these endpoints?
afaik we still don't have an auth strategy for mods (which wouldn't need these) and the site itself doesn't need them.

Including them in the swagger docs is also questionable as they're authenticated endpoints (only reason is for us to point mod authors at, but we can't do that yet as no auth)

Pixelguymm: I thought these endpoints would need to be public so they could be used by BeatSync.

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    post<BookmarksApi.Add, BookmarkRequest>("Add a bookmark".responds(ok<BookmarkUpdateResponse>())) { _, req ->
+        call.response.header("Access-Control-Allow-Origin", "*")
+
+        requireAuthorization { sess ->
+            val mapId = req.mapId
+
+            val added = transaction {
+                val exists = isBookMarked(mapId, sess.userId)
```
Top-Cat, line 85: Why check when we can just insert ignore?

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
```diff
@@ -0,0 +1,157 @@
...
+
+    get<BookmarksApi.Bookmarks>("Get bookmarks from the logged-in user".responds(ok<BookmarkResponse>())) {
+        call.response.header("Access-Control-Allow-Origin", "*")
+
+        requireAuthorization { sess ->
+            val maps = transaction {
+                Playlist
+                    .joinMaps()
+                    .slice(Beatmap.columns)
+                    .select { Playlist.id eq getBookmarksId(sess.userId) }
```
Top-Cat, line 146: This could be made a join instead of two queries

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
```diff
@@ -0,0 +1,157 @@
...
+        User.update({ User.id eq userId }) {
+            it[bookmarksId] = id
+        }
+    }
+
+    return id
+}
+
+fun isBookMarked(mapId: Int, userId: Int): Boolean {
+    val bookmarksId = getBookmarksId(userId)
```
Top-Cat, line 65: Again, could be a join

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
```diff
@@ -0,0 +1,157 @@
...
+        call.response.header("Access-Control-Allow-Origin", "*")
+
+        requireAuthorization { sess ->
+            val mapId = req.mapId
+
+            val removed = transaction {
+                val exists = isBookMarked(mapId, sess.userId)
+
+                if (exists) {
+                    val bookmarksId = getBookmarksId(sess.userId)
```
Top-Cat, line 119: Subquery, exists check is not needed.

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
Pixelguymm: Because I didn't know that existed :).

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
```diff
@@ -0,0 +1,158 @@
...
+        it[description] = "Maps you've bookmarked, automatically updated."
+        it[owner] = userId
+    }?.value?.also { pl ->
+        User.update({ User.id eq userId }) {
+            it[bookmarksId] = pl
+        }
+    }
+}
+
+fun isBookmarked(mapId: Int, userId: Int): Boolean {
```
Top-Cat, line 62: This function needs to go too, you can't run 20 queries for every search result

src/commonMain/resources/db/migration/V41__BookmarksIndex.sql
```diff
@@ -0,0 +1,5 @@
+CREATE INDEX IF NOT EXISTS playlist_type
+    ON public.playlist USING btree
+        ("createdAt" ASC NULLS LAST, "type")
+    TABLESPACE pg_default
+    WHERE "type" != 'Private' AND "deletedAt" IS NULL;
```
Top-Cat, line 5: Need to add an index so that the insert ignore works:
```sql
CREATE UNIQUE INDEX IF NOT EXISTS bookmark_playlist
    ON public.playlist USING btree ("owner")
    WHERE "type" = 'System';
```

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
```diff
@@ -0,0 +1,151 @@
...
+import org.jetbrains.exposed.sql.insertIgnore
+import org.jetbrains.exposed.sql.insertIgnoreAndGetId
+import org.jetbrains.exposed.sql.select
+import org.jetbrains.exposed.sql.transactions.transaction
+import org.jetbrains.exposed.sql.update
+
+@Location("/api")
+class BookmarksApi {
+    @Group("Bookmarks")
+    @Location("/bookmarks/add")
```
Top-Cat, line 37: I think both playlist map changes and followers use a single endpoint with the add/remove being part of the request.
This would be ok but it's inconsistent.

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
Top-Cat: CORS headers only apply to browsers, beatsync doesn't need them.

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
Pixelguymm: Does the same go for the option endpoints?

src/jvmMain/kotlin/io/beatmaps/api/bookmark.kt
Pixelguymm: Does the same go for the option endpoints?

Top-Cat: Yes

Issue #13, opened 2021-08-06
[Request] Map length on front of the map, contributers tab?


3Stans:
Knowing how long a map is without having to open the map page will probably be helpful in the case I go to search new nekopara maps and get too hyped to check out the map page only to find out the nekopara map was only about 15 seconds long. Though personally it took me a bit to notice the one on the old beatsaver and I wouldn't mind a 15 second long nekopara map, would probably be useful for others from what I've heard

And a contributors tab would, I feel be nice for the cases where a mapper may put playtesters, other mappers/lighters in there but not in the map authors field.

helencarnate:
See Issue #7 for a mock that adds song duration (and other common screening data) to the Map Card.

Top-Cat:
Map length is now visible on the cards.

I would recommend not adding multiple suggestions to a single issue, if you want to flesh out your other suggestion please open a separate issue for it.

PR #178, opened 2023-01-14
Add action filter to modlog

Pixelguymm:
Requires beatmaps-io/beatsaver-common-mp#13


src/jsMain/kotlin/io/beatmaps/modlog/main.kt
```diff
@@ -63,20 +69,26 @@ class ModLog : RComponent<ModLogProps, ModLogState>() {
...
                 this.mod = mod
                 this.user = user
+                this.type = type
                 resultsKey = Any()
             }
         }
+
+        if (type != state.type) {
+            typeRef.current?.value = type?.name ?: ""
+        }
```
Top-Cat, line 91: This seems redundant as you already set the value above

Issue #14, opened 2021-08-06
Redirect old mapper profile links


Top-Cat:
For example: https://beatsaver.com/uploader/5cff0b7698cc5a672c85479f

Top-Cat:
Done

PR #179, opened 2023-01-18
Add review deletion alert

Pixelguymm:
Requires beatmaps-io/beatsaver-common-mp#14


src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -294,6 +296,14 @@ fun Route.reviewRoute() {
                         ReviewDeleteData(deleteReview.reason, info.first, info.second),
                         single.userId
                     )
+
+                    Alert.insert(
+                        "Your review was deleted",
+                        "A moderator deleted your review on #$mapId.\n" +
```
Top-Cat, line 302: You don't want the integer map id

PR #186, opened 2023-05-10
update rating system to weigh vote ratio more heavily than currently …

Semantic-spaces:
…and weigh raw number of up votes less


Top-Cat:
This will only apply to maps when they receive new votes

Semantic-spaces:
Thanks for the update, will read through the codebase more to understand what I need to change

Semantic-spaces:
Righto, I think this should do the trick! Pixelguy helped with getting the sql correct & tested to make sure the results are as expected.

PR #187, opened 2023-05-16
Use ArcViewer for map previews

AllPoland:
Replaces BSViewer links and embeds with ArcViewer ones

### DIFF
--- a/src/jsMain/kotlin/io/beatmaps/index/preview.kt
+++ b/src/jsMain/kotlin/io/beatmaps/index/preview.kt
@@ -22,7 +22,7 @@ import react.dom.h5
 import react.dom.iframe
 import react.setState
 
-const val previewBaseUrl = "https://skystudioapps.com/bs-viewer/"
+const val previewBaseUrl = "https://allpoland.github.io/ArcViewer/"
 
 external interface ModalState : State {
     var modal: ModalData?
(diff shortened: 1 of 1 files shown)


src/jsMain/kotlin/io/beatmaps/index/preview.kt
```diff
@@ -36,7 +36,7 @@ class ModalComponent : RComponent<Props, ModalState>() {
     private val modal = createRef<HTMLDivElement>()
     private val iframe = createRef<HTMLIFrameElement>()
 
-    fun show(downloadURL: String) = showIframe("$previewBaseUrl?noProxy=true&url=$downloadURL")
```
Top-Cat, line 39: Do you have functionality in arcviewer for disabling proxying?

Tried to load the previewer locally but it fails to load `https://cors.bsmg.dev/http://localhost:8080/cdn/######.zip`

AllPoland: No, I've been disabling proxy based on a whitelist, but I can add it as a URL argument

AllPoland: I've just added the functionality to ArcViewer and updated the PR, should work now.

PR #190, opened 2023-06-08
Lower minimum reviews required to display review sentiment

XAce1337manX:
Discord suggestion.
image


Top-Cat:
Don't know about this. Review sentiment doesn't actually do much so importance is low but a lower threshold surely just enables small groups leaving negative reviews.

PR #191, opened 2023-06-12
Show link to BL leaderboard even if there are no scores

NSGolova:


Top-Cat:
I don't exactly know what you're doing in the backend here but I seem to be able to use this endpoint to create leaderboards for characteristics that don't exist which then show up on the beatleader website to pick from.

Doesn't work for difficulties that aren't real though. Is this intended? Maybe for mods generating extra characteristics?

NSGolova:
Yes, it lazily creates leaderboards on request. For both existing in the map characteristics and custom (Chirality or Controllable mods for example). There is no whitelist for the characteristic name, so it's potentially abusable, but I can clean them later if needed. Do you want to show these characteristics on Saver as well?

Top-Cat:
Not really. Think I'll stick to showing information about the map rather than things that _could_ exist in modded play

PR #193, opened 2023-06-29
Collaboration

Pixelguymm:
Completely open to feedback on anything, this is just one implementation that I had in mind.


Top-Cat:
Unsure that this system should be using alerts. There's several changes here that are specific to making collaboration requests work that we don't use for "normal" alerts. An extra column specifically to link the request, making alerts deletable and because a collaboration row is a single user they'd be 1-to-1 rather than the 1-to-many design that alerts currently has.

I don't think we want a second alerting system for it (current one needs love as it is) but maybe stick with the current pattern of linking from the alert to another page where we present the request (could be map specific rather than request specific).

Pixelguymm:
From a user's perspective, I think working with alerts is the most intuitive way to go about it. I agree that the current implementation is kind of hack-y though. 

The added column could probably be done differently, though it does come with the benefit that alerts from cancelled or accepted/rejected invites can easily be removed. We wouldn't want to have users get spammed with alerts that will stick around afterwards. Another way to go about that might be to generate alerts from unanswered invites instead. The alerts don't contain any information that we can't easily get from the invite.

Deleting alerts could also be refined, I'm not too proud of [this bit](https://github.com/Pixelguymm/beatsaver-main/blob/d3e8a828994ce7b63e770d68f28dfcedcb273e69/src/jsMain/kotlin/io/beatmaps/user/alerts/alert.kt#L99C12-L99C12)...

Top-Cat:
> Deleting alerts could also be refined, I'm not too proud of [this bit](https://github.com/Pixelguymm/beatsaver-main/blob/d3e8a828994ce7b63e770d68f28dfcedcb273e69/src/jsMain/kotlin/io/beatmaps/user/alerts/alert.kt#L99C12-L99C12)...

Yeah, I found that. I'm yet to actually pull down and work out what that does.

> We wouldn't want to have users get spammed with alerts that will stick around afterwards.

Isn't that currently what happens though? What's the difference between read and deleted? I guess I'm not sure when I'd choose to use one over the other.

> From a user's perspective, I think working with alerts is the most intuitive way to go about it. I agree that the current implementation is kind of hack-y though.

I'm not saying we shouldn't be using the alerts on the front end, it's more the implementation I wasn't sure of.

---

Another note: What about a new search keyword that triggers searching collaborators. (Or a new one that backports the current functionality) in case people want to search for maps on an account rather than someone has worked on

Pixelguymm:
> Yeah, I found that. I'm yet to actually pull down and work out what that does.

The data returned from the API when marking the alert still includes the alert that would be deleted right after. So... I manually remove it.

> Isn't that currently what happens though? What's the difference between read and deleted? I guess I'm not sure when I'd choose to use one over the other.

Deleting would mostly be useful for canceling or rejecting requests. It prevents invalid requests from sticking around, either as an unread or read alert, when cancelling or rejecting respectively.

> I'm not saying we shouldn't be using the alerts on the front end, it's more the implementation I wasn't sure of.

What are your thoughts on what I proposed here:

> Another way to go about that might be to generate alerts from unanswered invites instead. The alerts don't contain any information that we can't easily get from the invite.

---

> Another note: What about a new search keyword that triggers searching collaborators. (Or a new one that backports the current functionality) in case people want to search for maps on an account rather than someone has worked on

I could exclude collaborators from the `mapper:` search. I'd also have to change the way maps are fetched on profile pages then, I think the `/maps/uploader` page will return the same results.

Top-Cat:
> What are your thoughts on what I proposed here:
> 
> > Another way to go about that might be to generate alerts from unanswered invites instead. The alerts don't contain any information that we can't easily get from the invite.

Not sure I understood initially but I think yes.

--------

Have you thought much about abuse? Some kid inviting you as a collaborator on all their maps. And if you reject the invite they can invite you again.

Pixelguymm:
Haven't thought about that much, other than the fact that you can't spam someone with requests from the same map.
The least I can do is make sure suspended accounts can't invite anyone. I'll make that a priority.

src/commonMain/resources/db/migration/V49__AlertCascade.sql
Top-Cat, line 1: Is this migration still necessary?

Top-Cat:
> Haven't thought about that much, other than the fact that you can't spam someone with requests from the same map. The least I can do is make sure suspended accounts can't invite anyone. I'll make that a priority.

Wondering if an option for users to opt out of receiving collaboration requests would be helpful. I think we can see how we go though and look to add that later if there are issues.

Top-Cat:
Also there's a conflict

src/jvmMain/kotlin/io/beatmaps/api/maps.kt
```diff
@@ -519,13 +524,16 @@ fun Route.mapDetailRoute() {
...
+                .joinCollaborators()
                 .select {
                     Beatmap.id.inSubQuery(
                         Beatmap
                             .joinVersions()
+                            .joinCollaborations()
                             .slice(Beatmap.id)
                             .select {
-                                Beatmap.uploader.eq(it.id) and (Beatmap.deletedAt.isNull())
+                                (Beatmap.uploader eq it.id or (Collaboration.collaboratorId eq it.id)) and
```
Top-Cat, line 535: As per my previous comments about the `mapper:` tag in search. I think it would be better to either have a parameter on this endpoint or a new endpoint that enabled bringing in collaborators.

src/commonMain/resources/db/migration/V49__AlertCascade.sql
Pixelguymm: It's not. I can delete it if you want, but I don't see harm in keeping it either.

Pixelguymm:
> Wondering if an option for users to opt out of receiving collaboration requests would be helpful. I think we can see how we go though and look to add that later if there are issues.
I think this would come naturally with adding alert settings, which I should probably get started on soon.

src/jvmMain/kotlin/io/beatmaps/api/alerts.kt
```diff
@@ -47,26 +51,46 @@ import org.jetbrains.exposed.sql.update
...
     .select {
-        (AlertRecipient.recipientId eq userId) and
-            AlertRecipient.readAt.isNull()
+        (AlertRecipient.recipientId eq userId and AlertRecipient.readAt.isNull()) or
+                (Collaboration.collaboratorId eq userId and not(Collaboration.accepted))
     }.count().toInt()
 
+fun UserAlert.Companion.from(collaboration: CollaborationDAO) = UserAlert(
+    head = "Collaboration Proposal",
+    body = "You have been invited to be a collaborator on #${collaboration.mapId}. " +
```
Top-Cat, line 62: `mapId` is an int here, needs to be wrapped with `toHexString`

src/jsMain/kotlin/io/beatmaps/maps/mapinfo.kt
```diff
@@ -147,6 +154,143 @@ val tagPicker = fc<TagPickerProps> { props ->
...
+                            }
+                        }
+                        +"Search"
+                    }
+                }
+
+                state.foundUsers?.filter {
+                    it.id != userData?.userId && state.collaborators?.none { c ->
+                        c.collaborator.id == it.id
+                    } ?: true
```
Top-Cat, line 250: I think != false is more idiomatic than ?: true

Top-Cat:
Lastly...
```
src\jsMain\kotlin\io\beatmaps\user\alerts\main.kt:196:1 Unexpected indentation (36) (should be 32)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:58:1 Unexpected indentation (16) (should be 12)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:64:1 Unexpected indentation (12) (should be 8)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:65:1 Unexpected indentation (12) (should be 8)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:77:18 Missing newline after "("
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:78:1 Unexpected indentation (24) (should be 20)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:78:121 Missing newline before ")"
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:79:1 Unexpected indentation (24) (should be 20)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:80:1 Unexpected indentation (28) (should be 24)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:81:1 Unexpected indentation (28) (should be 24)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:82:1 Unexpected indentation (28) (should be 24)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:83:1 Unexpected indentation (24) (should be 20)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:120:1 Unexpected indentation (20) (should be 16)
src\jvmMain\kotlin\io\beatmaps\api\alerts.kt:131:13 Unexpected newline before "else"
src\jvmMain\kotlin\io\beatmaps\api\maps.kt:538:34 Missing newline after "("
src\jvmMain\kotlin\io\beatmaps\api\maps.kt:539:1 Unexpected indentation (40) (should be 36)
src\jvmMain\kotlin\io\beatmaps\api\maps.kt:540:1 Unexpected indentation (40) (should be 36)
src\jvmMain\kotlin\io\beatmaps\api\maps.kt:541:1 Unexpected indentation (32) (should be 36)
src\jvmMain\kotlin\io\beatmaps\api\maps.kt:541:33 Missing newline before ")"
src\jvmMain\kotlin\io\beatmaps\api\search.kt:22:1 Unused import
src\jvmMain\kotlin\io\beatmaps\api\search.kt:31:1 Unused import
src\jvmMain\kotlin\io\beatmaps\api\search.kt:216:28 Missing newline after "("
src\jvmMain\kotlin\io\beatmaps\api\search.kt:217:1 Unexpected indentation (28) (should be 24)
src\jvmMain\kotlin\io\beatmaps\api\search.kt:217:29 Line must not begin with "+"
src\jvmMain\kotlin\io\beatmaps\api\search.kt:218:1 Unexpected indentation (28) (should be 24)
src\jvmMain\kotlin\io\beatmaps\api\search.kt:218:29 Line must not begin with "+"
src\jvmMain\kotlin\io\beatmaps\api\search.kt:218:97 Missing newline before ")"
src\jvmMain\kotlin\io\beatmaps\api\users.kt:1223:1 Unexpected blank line(s) before "}"
src\jvmMain\kotlin\io\beatmaps\api\collaboration.kt:94:1 Unexpected indentation (24) (should be 20)
src\jvmMain\kotlin\io\beatmaps\api\collaboration.kt:95:1 Unexpected indentation (28) (should be 20)
src\jvmMain\kotlin\io\beatmaps\api\collaboration.kt:96:1 Unexpected indentation (24) (should be 16)
src\jvmMain\kotlin\io\beatmaps\api\collaboration.kt:121:1 Unexpected blank line(s) before "}"
src\jvmMain\kotlin\io\beatmaps\util\ownership.kt:1:1 File must end with a newline (\n)
```

PR #195, opened 2023-07-09
Add reviews tab to account pages

Pixelguymm:


src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -152,6 +152,7 @@ fun Route.reviewRoute() {
         val reviews = transaction {
             try {
                 Review
+                    .join(reviewerAlias, JoinType.INNER, Review.userId, reviewerAlias[User.id])
```
Top-Cat, line 155: Pretty sure this was purposefully omitted. The reviewer is always the user. You should already have that information on the profile page.

src/jsMain/kotlin/io/beatmaps/shared/sharedComponents.kt
```diff
@@ -216,6 +215,31 @@ val playlistOwner = fc<PlaylistOwnerProps> { props ->
     }
 }
 
+external interface ReviewerProps : Props {
```
Top-Cat, line 218: Am I missing the reason this wasn't just a new property on the existing component?

Pixelguymm: I figured splitting them would be cleaner. You'd never use that property on playlists, or anywhere other than reviews. I can merge them again if you want though, just say the word.

Top-Cat: It's not a big deal, just results in duplicating the rest of the code. Probably a small enough component to not worry about too much.

src/jsMain/kotlin/io/beatmaps/shared/review/main.kt
```diff
@@ -94,8 +95,10 @@ class ReviewTable : RComponent<ReviewTableProps, ReviewTableState>() {
                 attrs.container = resultsTable
                 attrs.renderElement = InfiniteScrollElementRenderer { rv ->
                     reviewItem {
-                        obj = rv
-                        userId = rv?.creator?.id ?: -1
+                        obj = rv?.apply {
```
Top-Cat, line 98: I think this should probably use copy rather than apply to keep the property a val

PR #204, opened 2023-09-16
Allow admins to remove contributors

Pixelguymm:


Top-Cat:
Should use the isAdmin helper function, at least when doing updates

Top-Cat:
ok to test

PR #251, opened 2023-11-10
Add audio previews to map pages

Pixelguymm:


src/jsMain/kotlin/io/beatmaps/util/audio.kt
```diff
@@ -8,10 +8,10 @@ fun useAudio() = useRef(
     Audio().also {
         it.volume = 0.4
     }
-).also { audio ->
+).also {
     useEffectOnce {
         cleanup {
-            audio.current?.pause()
+            it.current?.pause()
```
Top-Cat, line 14: Any reason?

src/jsMain/kotlin/io/beatmaps/shared/audioPreview.kt
```diff
@@ -0,0 +1,121 @@
...
+                    it.pause()
+                    it.currentTime = 0.0
+                }
+            }
+        }
+    }
+
+    override fun RBuilder.render() {
+        div("audio-progress") {
+            attrs.jsStyle["height"] = props.size + "px"
```
Top-Cat, line 96: Clearer as a block imo
```
attrs.jsStyle {
    height = props.size + "px"
    width = props.size + "px"
}
```

Top-Cat:
Not sure this is quite right?
Screenshot 2023-11-11 at 13 42 32

src/jsMain/kotlin/io/beatmaps/util/audio.kt
Pixelguymm: Moved it to a file where it shadowed another variable. Then moved it back and forgot to change it. I'll revert that.

Pixelguymm:
> Not sure this is quite right?

I couldn't get uploads to work locally, despite having all the folders. That's the one part I couldn't test -_-.

Issue #41, opened 2021-09-22
If a user clicks "Change username" without changing the username value, the request will still go through and put the username change on cooldown


namaki-mono:
When clicking on "Change username" in the user profile's "Account" section (https://beatsaver.com/profile#account) when the username field is unchanged/untouched (i.e. not dirty, still exactly the same as the user's current account name), the request will go through successfully and put the username change on the 24-hour cooldown. 

Based on the code here: 
https://github.com/beatmaps-io/beatsaver-main/blob/66cd4e0c8a552709f6879be9306232a295a47410/src/jvmMain/kotlin/io/beatmaps/api/users.kt#L168
Clicking "Change username" with the same name should return the error "Username already taken", but the same username is being caught as a unique username.

Error received after clicking on "Change username" twice:
![image](https://user-images.githubusercontent.com/81204441/134320725-d7636db7-f90f-4783-aa3d-bdf64c2ce47c.png)

Top-Cat:
Thank you Mr Sloth, there's a new message on the client side and if you somehow bypass that it shouldn't count as a reset if your name doesn't change

PR #256, opened 2023-11-29
Bot map redesign

Pixelguymm:
Requires beatmaps-io/beatsaver-common-mp#41

### DIFF
--- a/src/commonMain/resources/db/migration/V63__BotMapRedesign.sql
+++ b/src/commonMain/resources/db/migration/V63__BotMapRedesign.sql
@@ -1,0 +1,18 @@
+CREATE TYPE public.aiDeclarationType AS ENUM
+(
+    'Admin',
+    'Uploader',
+    'SageScore',
+    'None'
+);
+
+ALTER TABLE beatmap
+    ADD "declaredAi" aiDeclarationType;
+
+UPDATE beatmap
+SET "declaredAi" = 'SageScore'
+FROM beatmap AS b
+         LEFT JOIN versions AS v ON b."mapId" = v."mapId"
+WHERE beatmap."mapId" = b."mapId"
+  AND v.state = 'Published'
+  AND v."sageScore" < 0;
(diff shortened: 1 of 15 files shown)


Pixelguymm:
The migration only takes "sageScore" field into account, not "ai" or "automapper". I figured it wouldn't be too bad to have a reset of sorts, but I can include those two fields if needed.

Top-Cat:
Need a few days to run this and double check how the UI looks but this seems pretty good

src/jvmMain/kotlin/io/beatmaps/api/maps.kt
```diff
@@ -263,26 +265,29 @@ fun Route.mapDetailRoute() {
...
+            val result = transaction {
+                if (!user.isAdmin() && !(isUploader(mapUpdate.id, user.userId) && mapUpdate.automapper)) {
+                    false
+                } else {
                     Beatmap.update({
                         (Beatmap.id eq mapUpdate.id)
                     }) {
-                        it[automapper] = mapUpdate.automapper
+                        it[declaredAi] = when {
+                            !mapUpdate.automapper && user.isAdmin() -> AiDeclarationType.None
```
Top-Cat, line 280: The `user.isAdmin()` check here isn't necessary as to be in this block it's either true or automapper isn't false

However I would probably rewrite this whole block as below to avoid doing a separate query to get the uploader:
```kotlin
val admin = user.isAdmin()
Beatmap.update({
    (Beatmap.id eq mapUpdate.id).let { q ->
        if (user.isAdmin()) {
            q // If current user is admin don't check the user
        } else {
            q and (Beatmap.uploader eq user.userId)
        }
    }
}) {
    it[declaredAi] = when {
        !mapUpdate.automapper && admin -> AiDeclarationType.None
        mapUpdate.automapper && admin -> AiDeclarationType.Admin
        else -> AiDeclarationType.Uploader
    }
    it[updatedAt] = NowExpression(updatedAt.columnType)
} > 0
```

The only situation this would have a different affect is when the uploader makes a call that tries to set automapper to false. Instead of doing nothing this would treat that as true which seems like a non-issue.

src/commonMain/resources/db/migration/V63__BotMapRedesign.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TYPE public.aiDeclarationType AS ENUM
+(
+    'Admin',
+    'Uploader',
+    'SageScore',
+    'None'
+);
+
+ALTER TABLE beatmap
+    ADD "declaredAi" aiDeclarationType;
```
Top-Cat, line 10: Should this have
```sql
NOT NULL DEFAULT 'None';
```

The code doesn't mark this as nullable but the default it null and nothing shows up for me locally.

PR #258, opened 2023-12-08
Use docker volumes instead of local filesystem bind mounts

JoelLarson:
This PR addresses an issue reported in the Discord. When using Windows Subsystem for Linux, local directories being used for file-systems mounted to the docker container will have permission problems.

This change is not backwards compatible, but should be worth the change since the DB is rebuilt easily. In the related Docker Volume documentation, the backup/restore can be used if data is still important for individuals.

Related Resources:

* [Docker Compose Volumes](https://docs.docker.com/compose/compose-file/07-volumes/)
* [Docker Volumes](https://docs.docker.com/storage/volumes/)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     ports:
       - "5432:5432"
     volumes:
-      - ./db:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


Top-Cat:
test this please

- For jenkins

PR #261, opened 2023-12-15
Update .editorconfig to enforce single name imports

Pixelguymm:


.editorconfig
```diff
@@ -3,3 +3,8 @@ ktlint_standard_filename = disabled
 ktlint_standard_argument-list-wrapping = disabled
 ij_kotlin_allow_trailing_comma = false
 ij_kotlin_allow_trailing_comma_on_call_site = false
+ij_kotlin_import_nested_classes = false
+ij_kotlin_imports_layout = *,java.**,javax.**,kotlin.**,^
+ij_kotlin_name_count_to_use_star_import = 2147483647
+ij_kotlin_name_count_to_use_star_import_for_members = 2147483647
+ij_kotlin_packages_to_use_import_on_demand =
```
Top-Cat, line 10: Not that this file gets checked but it should still end in a newline

Top-Cat:
Also ij_kotlin_packages_to_use_import_on_demand seems to completely break ktlint

PR #262, opened 2023-12-16
Use human tag names when showing the selected tags in general search

JoelLarson:
This PR will switch the filters for tags to use human labels instead of the key.

image

I'm not a fan of this solution. I think the filters should have an interface that defines whether the filter is "inverse", and has the human and key on it. Then the code could be reduced to:

```kotlin
filters.map { it.humanFormatted }.joinToString(", ")
```

I've been having a hard time figuring out how the `!` ends up in the output, and how I could generalize the filters and tags.

### DIFF
--- a/src/jsMain/kotlin/io/beatmaps/shared/search/search.kt
+++ b/src/jsMain/kotlin/io/beatmaps/shared/search/search.kt
@@ -5,7 +5,6 @@ import external.Moment
 import io.beatmaps.common.MapTag
 import io.beatmaps.common.SearchOrder
 import io.beatmaps.common.SortOrderTarget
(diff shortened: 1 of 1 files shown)


Pixelguymm:
This is a small thing, but:
```kt
filters.map { /* etc */ }.joinToString(", ")
```
can be shortened to:
```kt
filters.joinToString(", ") { /* etc */ }
```

Top-Cat:
> I think the filters should have an interface that defines whether the filter is "inverse", and has the human and key on it.

I don't know what you mean.

I'm not really a fan of this, the intent was to show shorthand information as there is limited space. Other fields are shortened "automapper" -> "ai" etc.

Issue #43, opened 2021-09-26
Add a data point to the site where you can include other mappers' accounts as collaborators


stillinkansas:
Having a way to specify collaborators by account would be pretty nice.

If I work on a map with someone and they upload it to their account, having that map show up on my map list too even though I didn't upload it would be cool.

If this were to be a feature, I think you would need to add another field to the map cards in the search list that differentiates between uploaders and collaborators/mappers or at least specifies the uploader field.

helencarnate:
> Having a way to specify collaborators by account would be pretty nice.
This is handled on Bsaber by manually adding multiple mapper tags. Might be worth discussing a coordinated way to manage as part of our overall community site conversation (that I still need to schedule)

3Stans:
Here are my thoughts on how this could work (pasted from my closed request) ;


_**How would this work?**_ 
My idea of this would be that after a couple, or a group of mappers have finished collaborating on a map, they can decide who gets to upload it, and the person who gets to upload will be the host of the map who can then invite the other participants and allow them to link the map on their account. The linked map would appear the same as the host map and any other map on the participants account but with any votes going to the host map. 

It could be used to promote maps as well but this is not what I intend the purpose of the feature to be.

The feature could also possibly add an extra little box in the map page with a list of collaborators which are clickable and take you to the pages of each mapper.

_**Okay.... How will it work on the participant's page?**_
The host will have the main controls of the map with just a few extra buttons. If members of the collab wish to have the map on their page, then they do the following:

1. Click an addition "Invite Participant" button (or something along that line)
2. Search for participant mapper name
3. Invite

On the participants side, they have it a little easier and just have to check their alerts tab for an invite. If they have an invite then they have an option to accept or decline, if they choose to accept then the map will appear as the latest upload on their profile but will not re-post to the latest feed on BeatSaver. 

### - What about moderation of this feature?**

- Well first off, participants have the option to accept and decline the invite, declining the invite will simply clear the notif from the alerts tab and the map will show on their account.
- Participants will also have no control over the map on their account, they cannot edit, unpublish, republish, or delete the map from 
BeatSaver, they can only remove the linked map from _their_ account.
- People can also not request to be invited to reduce spam of some kid who wants to try and pass off as helping with the creation of their favorite mod map Up and Down.
- Map hosts can also choose to revoke map linkage by going to the participants page with the linked map and clicking a newly added button labeled "Unlink Map"
[...]

PR #264, opened 2023-12-16
Review opt-out

nttfiles:
This PR will add a toggle to the user settings, allowing users to opt out of the reviews feature. This means no reviews can be left on any of the maps the user uploads, and existing reviews will be hidden. This setting can not be toggled on a map-by-map basis, it's all or nothing.

The current implementation does allow reviews on maps that have these users as collaborators, as long as the main uploader has reviews enabled. Whether or not this should the case would require further thought.

Some screenshots (don't mind the outdated UI elements):
image
image

Requires beatmaps-io/beatsaver-common-mp#43

There wasn't a clear consensus on whether or not people wanted this, but it's here if they do.

### DIFF
--- a/src/commonMain/kotlin/io/beatmaps/api/users.kt
+++ b/src/commonMain/kotlin/io/beatmaps/api/users.kt
@@ -39,7 +39,8 @@ data class UserDetail(
     val verifiedMapper: Boolean = false,
     val suspendedAt: Instant? = null,
     val playlistUrl: String? = null,
-    val patreon: PatreonTier? = null
+    val patreon: PatreonTier? = null,
+    val reviewsEnabled: Boolean = true
 ) {
     fun profileLink(tab: String? = null, absolute: Boolean = false) = UserDetailHelper.profileLink(this, tab, absolute)
     companion object
@@ -76,6 +77,9 @@ data class UserDiffStats(val total: Int, val easy: Int, val normal: Int, val har
 
 @Serializable
 data class AccountDetailReq(val textContent: String)
+
+@Serializable
+data class ReviewOptRequest(val enableReviews: Boolean)
 
 @Serializable
 data class RegisterRequest(val captcha: String, val username: String, val email: String, val password: String, val password2: String)
--- a/src/commonMain/resources/db/migration/V65__ReviewOptOut.sql
+++ b/src/commonMain/resources/db/migration/V65__ReviewOptOut.sql
@@ -1,0 +1,2 @@
+ALTER TABLE uploader
+    ADD "reviewsEnabled" bool not null default true;
--- a/src/jsMain/kotlin/io/beatmaps/maps/infotable.kt
+++ b/src/jsMain/kotlin/io/beatmaps/maps/infotable.kt
@@ -158,7 +158,7 @@ val infoTable = fc<InfoTableProps> { props ->
         props.map.stats.let { stats ->
             infoItem("Rating", "${stats.upvotes} / ${stats.downvotes} (${stats.scoreOneDP}%)")
 
-            if (ReviewConstants.COMMENTS_ENABLED) {
+            if (ReviewConstants.COMMENTS_ENABLED && props.map.uploader.reviewsEnabled) {
                 div(itemClasses) {
                     +"Reviews"
                     span("text-truncate ms-4") {
--- a/src/jsMain/kotlin/io/beatmaps/maps/main.kt
+++ b/src/jsMain/kotlin/io/beatmaps/maps/main.kt
@@ -138,7 +138,7 @@ val mapPage = fc<MapPageProps> { props ->
                     if (showComments) {
                         reviewTable {
                             attrs.map = it.id
-                            attrs.mapUploaderId = it.uploader.id
+                            attrs.uploader = it.uploader
                         }
                     } else if (version != null && it.deletedAt == null) {
                         scoreTable {
--- a/src/jsMain/kotlin/io/beatmaps/shared/review/main.kt
+++ b/src/jsMain/kotlin/io/beatmaps/shared/review/main.kt
@@ -15,6 +15,7 @@ import io.beatmaps.util.useDidUpdateEffect
 import org.w3c.dom.HTMLElement
 import react.Props
 import react.dom.div
+import react.dom.p
 import react.fc
 import react.useContext
 import react.useRef
@@ -22,7 +23,7 @@ import react.useState
 
 external interface ReviewTableProps : Props {
     var map: String?
-    var mapUploaderId: Int?
+    var uploader: UserDetail?
     var userDetail: UserDetail?
     var fullWidth: Boolean?
 }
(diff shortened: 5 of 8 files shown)


anothershad:
This PR is probably wildly out of date, but I'm still bumping this. Having seen plenty of mappers talk about wanting opt outs around Discord, I feel like I'm speaking for a good chunk of mappers that would like this.

Pixelguymm:
Don't worry about it being out of date. If enough people are behind this change, I'll make the needed adjustments.

Eileeks:
Bumping too, we absolutely need this option.

Top-Cat:
For anyone commenting can you describe the problems you see in reviews and how the approach taken in this PR specifically addresses those problems?

Pixelguymm:
I've edited the original PR message to include a more detailed description.

Eileeks:
@Top-Cat 
There are mappers that just wants to create and share their maps without having to deal with the pressure and the way too subjective opinions of other people in this community. Reviews enables unwanted opinions from anyone and can feel oppressive, plus they're eternally stuck on the associated map. 
The option of opting out of reviews give the liberty to mappers to decide if they want that or not.

TheCzar1994:
I agree with everything Skeelie said. In addition, just because a review meets TOS doesn't necessarily mean they're wanted. Also some people tend to insinuate that their reviews mean more or are better than others because of the attention they want/receive for doing it, which is counter productive. Personally, I'd rather have someone reach out to me privately to tell me they enjoyed a map than to have a review posted. That means a lot more to me. Again, just personal opinion though.

Top-Cat:
While that's all helpful context I don't feel like you've addressed how you feel this PR fixes those problems.

It's not that you need to convince me we need the feature. I've heard a lot of opinions on how it should be done (most more abusable imo than this) and more than likely something will be merged and everyone will keep complaining because it wasn't done right.

TimothyGillespie:
@Top-Cat 

I hope this comment addresses what you were looking for in regards to how this PR solves it. I did put some thought into it and try to see potential issues and different solutions opposed to just trying to convince you or anyone. Overall, the PR solutions looks very good and balanced to me. I'd only have one further suggestion for it.

**Problem with Reviews**:
I think that one is well iterated, let me perhaps put it differently, though:

They are a communication channel allowing natural language. This can discourage people greatly, but also conflicts with BeatSaver intend to be a site that is open and friendly to all, as it effectively provides no protection or agency to individuals who publish maps but are sensitive to words (but also doesn't seem great overall) - unless they don't participate on the platform.


**How this PR addresses the initial issue**:

This review allows disabling all natural language input towards the mapper (unless they are collaborator). They can now effectively close that communication channel.

It doesn't address the issue for collaborations, so a individual would need to restrict themselves in that form. This, however, seems reasonable and doesn't push them off the platform entirely. It's strictly an improvement even considering this in regard to the initial issue.

This PR disregards voting as a form of unwanted feedback which is likely felt as less strong as the expression is not in form of a language and more abstract.

**How this PR addresses follow up issues**:

**Abusability**: IIRC, some other implementation ideas had the issue of being able to suppress singular reviews by disabling it on a map by map basis. This PR addresses this by making it an all or nothing feature. You can still "target" a singular review, but you need to then surrender all other reviews on the account. You can't have your cake and eat it too, basically. 

The abusability could be minimized further by only not showing the reviews to the mapper and not notifying them. However, that seems to greatly lessen the effect of the "protection". The abusability with this PR seems rather low. I think the abusability of the review system is significantly greater and more impactful to the mapper than vice-versa, which is why I'd prefer the solution of the PR.

Loss of Data for Reviewer: If a Reviewer made a review they may want to go back to it personally. They no longer have access to it, though. I don't think this PR addresses this - it could show them on the mapper profile under the Review section when logged in as the mapper, but it would no longer publicly viewable.

Balancing Interest: This is a bit outside the scope of the technical, but there are three colliding interests that I see:

Mapper: Doesn't want to receive reviews or any communication
Reviewers: Wish to express their opinion or provide feedback
Players: Wish to read reviews and be inspired by reviews and not "waste" time by playing a bad map.

I would regard this under the purview of an uploaded, original map being a "gift" by the mapper to the general public. The Reviewer is entitled express their opinion, but the Reviewer is not entitled to be aided by BeatSaver for that to opinion to be transmitted to the Reviewer nor to be heard at all. The interests of the Player and the Community to find good maps via Reviews can be achieved by different means (Discord Chats, Curations, etc.) and they don't seem entitled to make demands on how a gift should be promoted. I don't think the gifter should be required to agree to reviews to save someones "time" either. The time of the player in magnitudes of a few minutes or less doesn't seem to outweigh the well-being of the mapper that gifted the map.

Undeceiver:
@TimothyGillespie I strongly disagree on many things but that's not why I'm here, as those discussions have already been had (and I'm not going to make a fuzz if this gets implemented).

However, I would appreciate if you would clarify the point:

> The abusability could be minimized further by only not showing the reviews to the mapper and not notifying them. However, that seems to greatly lessen the effect of the "protection". 

Could you explain how that would lessen the effect of the protection? It is very possible I don't see how it could happen, but the way I see it right now, about how the PR achieves the goal, you said:

> This review allows disabling all natural language input towards the mapper (unless they are collaborator). They can now effectively close that communication channel.

If the reviews are hidden just for the mapper (or anybody else who chooses to disable reviews from their system), this would still be 100% true, so why would that be lessened?

TimothyGillespie:
@Undeceiver I hope I understood your question correctly.

I would be lessened due to two reasons:

1. The review exists in first place. A review that doesn't exist cannot be perceived at all. The entire communication is cancelled and even some thoughts that would have arisen with the review won't arise on part of the reviewer even perhaps.
2. The review is still visible to the mapper if they are not signed in. Even if they are signed in they can still see it when looking in the Review Feed. Or if friend forwards it to them.

It would basically just be a way to "close your eyes" approach and they'd fully know that. Knowing that they are being talked about or might be talked about and they just don't know the details is not something that usually puts one's mind at ease. The communication happens, just without them.

That's why I say it's weaken protection of the uploader's peace of mind. Significantly, so I'd even say.

Going back to my quote:

> They can now effectively close that communication channel.

It wouldn't be totally closed anymore. The communication channel to them is "closed" on their end, but it's open on the other side now and still flows to the public.

Undeceiver:
> It wouldn't be totally closed anymore. The communication channel to them is "closed" on their end, but it's open on the other side now and still flows to the public.

Ultimately this is one of several things I disagree with the whole idea behind this. I strongly reject the notion that the mapper has the right to control what others say to others about their map. In my opinion, they can choose what they listen to, and giving them tools to not have to look at it is not something I strongly oppose (though I'd have other things to say about this based on other reasons why I don't agree, but I won't bring those up).

That said, again, it's not the biggest deal. I don't want to continue discussing here. You did answer my question, so that's fine. But I would like to point out that you have indicated that a part of the motivation here is not just about what the mapper sees, it's also about what other people talk about the map, regardless of the participation of the mapper.

TimothyGillespie:
> [...] motivation here is not just about what the mapper sees, it's also about what other people talk about the map, regardless of the participation of the mapper.

I don't think you mean it in that way, but no it's not about the the content of what other people talk about the map. It's the general expectation of talk and the fact that you can actually see it if it is public. BeatSaver, having this feature and Review Feed, encourages and promotes that.

Brings up perhaps another idea to address that, which is that Reviews are instead not pushed to the Review Feed and you need to be logged in (with an account different from the uploader) to see the review.

I'd personally still prefer the PR as it is, but that's a potential middle way that still addresses some concerns somewhat effectively.

PR #274, opened 2024-01-02
Followed Curations

Specture7:
Changes "Map Curation" to "Followed Curation" this can make it a little less confusing when you see two curation tabs in alerts


Top-Cat:
ok to test

PR #284, opened 2024-01-21
Redirect to uploader page after deleting a map

Pixelguymm:
As suggested by Bloodcloak


Top-Cat:
I think the original intention was:
> As a user I'd go back to my profile page when I delete my own map.


Should we be changing that journey? (And yes, public profile vs private matters ie deleting unpublished maps)

-----------------

> As an admin when deleting a map I want to go to the profile of the uploader of the map.

This makes sense to me, although I could also see:
> As an admin when deleting a map I want to stay on the same page and see the deleted state of the map.

From where you could click the link in the map info to go to the profile.

Pixelguymm:
The only time I've ever gone back to a deleted map's page is to go to the uploader afterwards, so I kept it as is for now. Behaviour when deleting unpublished maps should now work properly.

PR #285, opened 2024-01-24
Use BeatLeader accuracy instead of calculated percentage

Pixelguymm:
I couldn't really test this so I hope this is right.


Top-Cat:
scoresaber.kt needs updating to pass something (I guess null) into accuracy

PR #288, opened 2024-01-31
Text suggestions

Specture7:
This is a simple PR that has a few rewording suggestions


Top-Cat:
ok to test

Top-Cat:
Think it might be best to standardise the log naming. On user pages it's currently "Mod Log" but "ModLog" in the dropdown. Your changes have the same spacing inconsistency.

Maybe just "Mod Log" and "Review Log" everywhere?

Specture7:
Right, I can fix those

PR #297, opened 2024-02-10
Remove maps from playlist

Undeceiver:
https://discord.com/channels/882730837974609940/1185645147019554826/1185645147019554826


src/jsMain/kotlin/io/beatmaps/playlist/playlist.kt
```diff
@@ -315,15 +315,33 @@ val playlistPage = fc<Props> {
...
                             }
                         }
                     }
+                } else if (playlist?.owner?.id == userData?.userId){
+                    div("playlist") {
+                        maps.map {
+                            removeMapPlaylist {
+                                attrs.obj = it.map
+                                attrs.audio = audio
+                                attrs.playlistKey = playlist!!.playlistId
```
Top-Cat, line 337: Generally best to avoid `!!`
The condition `playlist?.owner?.id == userData?.userId` is actually true if you're logged out and the playlist on the page hasn't loaded yet (or doesn't exist)
In which case this will throw an exception.

It would be better to use `playlist != null && playlist.owner.id == userData?.userId`

src/jsMain/kotlin/io/beatmaps/playlist/playlist.kt
```diff
@@ -315,15 +315,33 @@ val playlistPage = fc<Props> {
...
+                                        attrs.mapId = it.map.id
+                                        attrs.removeMap = {
+                                            setMaps(maps - it)
+                                        }
                                     }
                                 }
                             }
                         }
                     }
+                } else if (playlist?.owner?.id == userData?.userId){
```
Top-Cat, line 331: Jenkins will be upset with the spacing here

Top-Cat (changes requested):

src/jsMain/kotlin/io/beatmaps/playlist/playlist.kt
Top-Cat: Actually it probably won't throw an exception because maps would be empty... still

Undeceiver: I had misunderstood that !! would somehow safely not execute the corresponding area if the null check failed. I'll change it.

src/jsMain/kotlin/io/beatmaps/playlist/playlist.kt
Undeceiver: Lol I am too, hadn't noticed.

Top-Cat (approved):

Top-Cat:
Probably shouldn't show up at all on search playlists. (The width is all wrong but it seems to be fine with normal playlists so I'm not worried about that)

Screenshot 2024-02-21 at 12 39 58

PR #299, opened 2024-02-21
Follow alert

Specture7:
Gives the option to disable and enable a new follower alert

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/api/users.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/users.kt
@@ -940,8 +940,9 @@ fun Route.userRoute() {
                             "@${user.uniqueName} is now following you!",
                             EAlertType.Follow,
                             req.userId
-                    )
-                    updateAlertCount(req.userId)
+                        )
+                        updateAlertCount(req.userId)
+                    }
                 }
             }
             call.respond(HttpStatusCode.OK)
--- a/src/jvmMain/kotlin/io/beatmaps/api/users.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/users.kt
@@ -46,6 +46,7 @@ import io.beatmaps.login.MongoSession
 import io.beatmaps.login.Session
 import io.beatmaps.login.cookieName
 import io.beatmaps.login.server.DBTokenStore
+import io.beatmaps.util.updateAlertCount
 import io.jsonwebtoken.Claims
 import io.jsonwebtoken.ExpiredJwtException
 import io.jsonwebtoken.Header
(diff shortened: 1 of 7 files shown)


Top-Cat:
ok to test

src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -919,21 +920,30 @@ fun Route.userRoute() {
...
             transaction {
-                if (req.following) {
-                    Follows.upsert(conflictIndex = Follows.link) { follow ->
-                        follow[userId] = req.userId
-                        follow[followerId] = user.userId
-                        follow[since] = NowExpression(since.columnType)
-                        follow[upload] = req.upload
-                        follow[curation] = req.curation
-                    }
+                val shouldAlert = Follows.select { Follows.userId eq req.userId }.empty()
```
Top-Cat, line 923: This condition checks if **anyone** follows the user in question

src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -919,21 +920,30 @@ fun Route.userRoute() {
...
+
+                Follows.upsert(conflictIndex = Follows.link) { follow ->
+                    follow[userId] = req.userId
+                    follow[followerId] = user.userId
+                    follow[since] = NowExpression(since.columnType)
+                    follow[upload] = req.upload
+                    follow[curation] = req.curation
+                    follow[following] = req.following
+                }
+                val followedUser = if (shouldAlert) {
```
Top-Cat, line 933: This nullable into null coalescing seems a bit too esoteric to me.
Just wrap the whole thing in the condition...

```kt
if (shouldAlert) {
    val followedUser = UserDao.wrapRow(User.select { User.id eq req.userId }.single())

    if (followedUser.followAlerts) {
        Alert.insert(
            "New Follower",
            "@${user.uniqueName} is now following you!",
            EAlertType.Follow,
            req.userId
        )
    }
}
```

(single should also be safe enough as the insert should fail if the user id is invalid)

src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -919,21 +920,30 @@ fun Route.userRoute() {
...
+                    User.select { User.id eq req.userId }.first().let { r -> UserDao.wrapRow(r) }
                 } else {
-                    Follows.deleteWhere {
-                        (Follows.userId eq req.userId) and (Follows.followerId eq user.userId)
-                    }
+                    null
+                }
+                if (followedUser?.followAlerts == true) {
+                    Alert.insert(
```
Top-Cat, line 939: I think you need to call `updateAlertCount(req.userId)` after Alert.insert

Specture7:
problems should be fixed now

Specture7:
re-added the bracket that was missing before

PR #304, opened 2024-03-14
Visualizing mod requirements

Specture7:
Shows mods like chroma, noodle, etc on a map that requires them


Top-Cat:
A screenshot of how this looks would be nice ;)

Ok to test

PR #309, opened 2024-03-26
Add og:author and og:author:url to Map and Playlist Page

TimothyGillespie:
Most PR seems to be pointing to main, so I copied that.
I did test it, couldn't test the embedding with the local setup though. The HTML Markup has been as expected though.

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
@@ -143,7 +143,7 @@ fun Route.mapController() {
 
                         // There can only be one URL so we only add it when there is no collaborator
                         // Otherwise, it may be confusing
-                        if (!it.collaborators.isNullOrEmpty()) {
+                        if (it.collaborators.isNullOrEmpty()) {
                             meta("og:author:url", "${Config.siteBase()}/profile/${it.uploader.id}")
                         }
                     }
--- a/src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
@@ -146,7 +146,6 @@ fun Route.mapController() {
                         if (it.collaborators == null || it.collaborators.size == 0) {
                             meta("og:author:url", "${Config.siteBase()}/profile/${it.uploader.id}")
                         }
-                        
                     }
                 } catch (_: NumberFormatException) {
                     // key isn't an int
--- a/src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
@@ -143,7 +143,7 @@ fun Route.mapController() {
 
                         // There can only be one URL so we only add it when there is no collaborator
                         // Otherwise, it may be confusing
-                        if (it.collaborators == null || it.collaborators.size == 0) {
+                        if (!it.collaborators.isNullOrEmpty()) {
                             meta("og:author:url", "${Config.siteBase()}/profile/${it.uploader.id}")
                         }
                     }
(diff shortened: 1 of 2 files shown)


Top-Cat:
Ok to test

src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
```diff
@@ -118,16 +119,40 @@ fun Route.mapController() {
...
+                                authors.addAll(collaborators.dropLast(1).map { it.name })
+                                authors.add("and ${collaborators.last().name}")
+                            }
+                        }
+                        val authorString = authors.joinToString(", ")
+                        meta("og:author", authorString)
+
+                        // There can only be one URL so we only add it when there is no collaborator
+                        // Otherwise, it may be confusing
+                        if (it.collaborators == null || it.collaborators.size == 0) {
```
Top-Cat, line 152: `if (it.collaborators.isNullOrEmpty()) {`

src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
```diff
@@ -118,16 +119,40 @@ fun Route.mapController() {
...
                         meta("og:image", it.publishedVersion()?.coverURL)
                         meta("og:description", it.description.take(400))
+                        
+                        // Joining mappers together for the og:author field so that:
+                        // 1. Uploader will be first
+                        // 2. All non-last collaborators will be joined with ", "
+                        // 3. The last collaborator will be joined with ", and "  
+                        val authors = mutableListOf(it.uploader.name)
+                        it.collaborators?.let { collaborators ->
+                            if (collaborators.isNotEmpty()) {
```
Top-Cat, line 142: `if (!it.collaborators.isNullOrEmpty()) {`

src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
```diff
@@ -118,16 +119,40 @@ fun Route.mapController() {
...
                         meta("og:site_name", "BeatSaver")
                         meta("og:title", it.name)
                         meta("og:url", "${Config.siteBase()}/maps/${it.id}")
                         meta("og:image", it.publishedVersion()?.coverURL)
                         meta("og:description", it.description.take(400))
+                        
+                        // Joining mappers together for the og:author field so that:
+                        // 1. Uploader will be first
+                        // 2. All non-last collaborators will be joined with ", "
+                        // 3. The last collaborator will be joined with ", and "
```
Top-Cat, line 139: The last item in a list shouldn't have the comma though

Top-Cat: Not sure if anyone wants to play code golf with me but here's an approach using reduce to allow this to be customised
```kotlin
val authors = (listOf(it.uploader) + (it.collaborators ?: listOf())).map { u -> u.name }.toMutableList()
val authorString = authors.reduceIndexed { index, acc, s -> "$acc${if (index == authors.lastIndex) " and" else ","} $s" }
```

Top-Cat:
Also there are 4 lint issues

TimothyGillespie:
For some reason I'm getting trolled by the local lint targets. But I managed now with the pipeline and local lint targets work now too. Probably, some silly mistake when using them on my part.

src/jvmMain/kotlin/io/beatmaps/controllers/map.kt
```diff
@@ -118,16 +119,33 @@ fun Route.mapController() {
...
+                        // 1. Uploader will be first
+                        // 2. All non-last collaborators will be joined with ", "
+                        // 3. The last collaborator will be joined with " and "
+                        val authors = (listOf(it.uploader) + it.collaborators.orEmpty()).map { u -> u.name }
+                        val authorString = authors.reduceIndexed { index, acc, s -> "$acc${if (index == authors.lastIndex) " and" else ","} $s" }
+                        meta("og:author", authorString)
+
+                        // There can only be one URL so we only add it when there is no collaborator
+                        // Otherwise, it may be confusing
+                        if (!it.collaborators.isNullOrEmpty()) {
```
Top-Cat: I think this is the wrong way round. The `!`, that was for the other if

TimothyGillespie: Yeah, just saw that as well. Already pushed. My bad!

PR #310, opened 2024-03-30
Notify followers when collaboration is accepted

Undeceiver:
Uses the same MapRelease notification type, but the message is slightly different, to account for the time that may pass between the map being published and the collaboration being accepted.

This is done in the same transaction that the collaboration is marked as accepted.

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
@@ -1,9 +1,14 @@
 package io.beatmaps.api
 
+import io.beatmaps.common.api.EAlertType
 import io.beatmaps.common.db.NowExpression
+import io.beatmaps.common.db.updateReturning
+import io.beatmaps.common.dbo.Alert
 import io.beatmaps.common.dbo.Beatmap
+import io.beatmaps.common.dbo.BeatmapDao
 import io.beatmaps.common.dbo.Collaboration
 import io.beatmaps.common.dbo.CollaborationDao
+import io.beatmaps.common.dbo.Follows
 import io.beatmaps.common.dbo.User
 import io.beatmaps.common.dbo.UserDao
 import io.beatmaps.common.dbo.collaboratorAlias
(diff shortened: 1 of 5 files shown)


Undeceiver:
image

src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
```diff
@@ -81,17 +85,59 @@ fun Route.collaborationRoute() {
...
+                                it[accepted] = true
+                            },
+                            Collaboration.mapId
+                        )
+                    } else {
+                        Collaboration.deleteWhere {
+                            Collaboration.id eq req.collaborationId and (Collaboration.collaboratorId eq sess.userId)
+                        }
+
+                        null
```
Top-Cat, line 104: Does this not mean that success is now always false for deletes (as mapId = null)

src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
```diff
@@ -81,17 +85,59 @@ fun Route.collaborationRoute() {
...
+                        row[Follows.followerId].value
+                    }
+
+                    val nonRecipients = Follows.select {
+                        Follows.userId eq map.uploaderId and Follows.following
+                    }.map { row ->
+                        row[Follows.followerId].value
+                    } + listOf(map.uploaderId.value)
+
+                    val finalRecipients = recipients - nonRecipients.toSet()
```
Top-Cat, line 129: Playing some more golf
It's somewhat complex but the database can quite efficiently do this filtering.

```kt
val followsAlias = Follows.alias("f2")
val recipients = Follows
    .join(followsAlias, JoinType.LEFT, followsAlias[Follows.followerId], Follows.followerId) {
        (followsAlias[Follows.userId] eq map.uploaderId) and followsAlias[Follows.following]
    }.select {
        followsAlias[Follows.id].isNull() and (Follows.followerId neq uploaderId) and
            (Follows.userId eq sess.userId) and Follows.upload and Follows.following
    }.map { row ->
        row[Follows.followerId].value
    }.toSet()
```

Top-Cat: As per discussion on discord these should at least get `.slice()`'d

Top-Cat:
Ok to test

PR #311, opened 2024-03-30
Bypass recaptcha check for oauth2 clients

TimothyGillespie:
We haven't discussed whether that's something we'd want to do, but if we do: Here would be the implementation for it.

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/api/playlist/create.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/playlist/create.kt
@@ -97,7 +97,11 @@ fun Route.playlistCreate() {
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


TimothyGillespie:
I think that lint job is stuck. It is running for 18 minutes now.

Top-Cat:
> I think that lint job is stuck. It is running for 18 minutes now.

What says it's been running for 18 minutes?

TimothyGillespie:
Actually, you're right. It doesn't say that. It might as well not be started which seems more likely. Same seems to be occurring on Undeceiver's PR.

Top-Cat:
This is not going to work. The oauth2 client id property of the session is used to bypass the manual confirmation on subsequent logins to the same oauth client and is therefore set on normal sessions.

PR #314, opened 2024-04-01
Fix link to Curators' curations tab from Map Card

TheCzar1994:
Should fix the link when clicking on a Curator's name from a Curated map card to go to their "curations" tab instead of "curated" as this URL doesn't exist.


Top-Cat:
Ok to test

PR #319, opened 2024-04-10
Following Filter

Specture7:
RePR from the last one

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/api/search.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/search.kt
@@ -20,6 +20,7 @@ import io.beatmaps.common.db.similar
 import io.beatmaps.common.db.unaccent
 import io.beatmaps.common.db.unaccentLiteral
 import io.beatmaps.common.db.wildcard
+import io.beatmaps.common.dbo.AccessTokenTable.slice
 import io.beatmaps.common.dbo.Beatmap
 import io.beatmaps.common.dbo.Difficulty
 import io.beatmaps.common.dbo.Follows
@@ -37,7 +38,7 @@ import io.beatmaps.common.dbo.joinVersions
 import io.beatmaps.common.toQuery
 import io.beatmaps.login.Session
 import io.beatmaps.util.cdnPrefix
-import io.beatmaps.util.requireAuthorization
+import io.beatmaps.util.optionalAuthorization
 import io.ktor.http.HttpStatusCode
 import io.ktor.server.application.call
 import io.ktor.server.locations.Location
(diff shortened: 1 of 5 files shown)


src/jvmMain/kotlin/io/beatmaps/api/search.kt
```diff
@@ -197,6 +200,16 @@ fun Route.searchRoute() {
         val sortArgs = searchInfo.sortArgsFor(actualSortOrder)
 
         newSuspendedTransaction {
+            val followingSubQuery = if (it.followed == true) {
+                requireAuthorization { _, user ->
```
Top-Cat, line 204: This shouldn't be a call to requireAuthorization, it very much seems that when not logged in you just ignore the property.
The call has the side effect of returning an error to the user when not logged in.

Just wrap the whole handler in optionalAuthorization and check if the user is null / auth type is None

Top-Cat: (Which will mean you won't need those changes to `requireAuthorization` and `optionalAuthorization`)

Top-Cat (changes requested):

Top-Cat:
But ok to test

Specture7:
Should be good now

src/jvmMain/kotlin/io/beatmaps/api/search.kt
```diff
@@ -187,108 +191,129 @@ fun Route.searchRoute() {
...
-                                        .notNull(it.curator) { o -> Beatmap.curator eq o }
-                                }
-                                .orderBy(*sortArgs)
-                                .limit(it.page)
+                if (searchInfo.escapedQuery != null && searchInfo.escapedQuery.startsWith("key:")) {
+                    Beatmap
+                        .slice(Beatmap.id)
+                        .select {
+                            Beatmap.id eq searchInfo.escapedQuery.substring(4)
+                                .toInt(16) and (Beatmap.deletedAt.isNull())
```
Top-Cat, line 218: That's a whacky place to add a linebreak (does it even need one?)

src/jvmMain/kotlin/io/beatmaps/api/search.kt
```diff
@@ -187,108 +191,129 @@ fun Route.searchRoute() {
...
+                        .select {
+                            Beatmap.id.inSubQuery(
+                                Beatmap
+                                    .joinUploader()
+                                    .crossJoin(
+                                        Versions
+                                            .let { q ->
+                                                if (needsDiff) {
+                                                    q.join(
+                                                        Difficulty,
```
Top-Cat, line 247: How small is your screen? 🙈

Specture7:
fixed the style a bit, i hope this one is good

PR #323, opened 2024-04-25
Fix cards not expanding when content overflows

Pixelguymm:


Top-Cat:
This is not the fix. There's javascript that is meant to cause them to animate out to full size.

PR #333, opened 2024-05-11
Notifications for map updates (+ option to prevent notifications when publishing)

Undeceiver:
(Different) notifications get sent to followers when updating a map (upon publishing a new version). But also, a new check input has been added to the Publish modal window (by default set to true) that allows the uploader to choose not to notify their followers (In case of very small updates, for example).

https://discord.com/channels/882730837974609940/1177359210233729044

image

image

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/util/scheduled.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/util/scheduled.kt
@@ -30,7 +30,12 @@ class CheckScheduled(private val rb: RabbitMQInstance) : TimerTask() {
                 ).mapNotNull {
                     schedulerLogger.info { "Scheduler publishing ${it.hash}" }
                     runBlocking { delay(1L) }
-                    if (publishVersion(it.mapId.value, it.hash, rb)) it else null
+                    // Undeceiver: I am not adding the "alert" parameter to the database because it feels like too much effort for
+                    // the unlikely case that someone scheduled a map for release but doesn't want to notify followers,
+                    // If this ever happens, the alert will go out regardless (if it's scheduled it's likely it's a big deal).
+                    // If we did want to consider the parameter, it can be passed here without any problem,
+                    // but then we'd have to save it in the database when scheduling. Can be done, not doing it now.
+                    if (publishVersion(it.mapId.value, it.hash, true, rb)) it else null
                 }
             }.forEach {
                 rb.publish("beatmaps", "maps.${it.mapId.value}.updated.state", null, it.mapId.value)
--- a/src/jvmMain/kotlin/io/beatmaps/util/scheduled.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/util/scheduled.kt
@@ -30,11 +30,6 @@ class CheckScheduled(private val rb: RabbitMQInstance) : TimerTask() {
                 ).mapNotNull {
                     schedulerLogger.info { "Scheduler publishing ${it.hash}" }
                     runBlocking { delay(1L) }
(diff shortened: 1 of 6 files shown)


src/jvmMain/kotlin/io/beatmaps/util/versions.kt
```diff
@@ -126,12 +128,22 @@ fun pushAlerts(map: BeatmapDao, rb: RabbitMQInstance?) {
...
+        )
+    } else {
+        Alert.insert(
+            "Map Updated",
+            "@${map.uploader.uniqueName} just updated #${toHexString(map.id.value)}: **${map.name}**.\n" +
+                "*\"${map.description.replace(Regex("\n+"), " ").take(100)}...\"*",
+            EAlertType.MapRelease,
+            recipients
+        )
+    }
```
Top-Cat, line 147: That's a lot of duplicated code just to switch out two small strings.

```kt
val (title, adjective) = if (map.lastPublishedAt == null) ("New Map Release" to "released") else ("Map Updated" to "updated")
Alert.insert(
    title,
    "@${map.uploader.uniqueName} just $adjective #${toHexString(map.id.value)}: **${map.name}**.\n" +
        "*\"${map.description.replace(Regex("\n+"), " ").take(100)}...\"*",
    EAlertType.MapRelease,
    recipients
)
```

src/jvmMain/kotlin/io/beatmaps/util/scheduled.kt
```diff
@@ -30,7 +30,12 @@ class CheckScheduled(private val rb: RabbitMQInstance) : TimerTask() {
                 ).mapNotNull {
                     schedulerLogger.info { "Scheduler publishing ${it.hash}" }
                     runBlocking { delay(1L) }
-                    if (publishVersion(it.mapId.value, it.hash, rb)) it else null
+                    // Undeceiver: I am not adding the "alert" parameter to the database because it feels like too much effort for
+                    // the unlikely case that someone scheduled a map for release but doesn't want to notify followers,
+                    // If this ever happens, the alert will go out regardless (if it's scheduled it's likely it's a big deal).
+                    // If we did want to consider the parameter, it can be passed here without any problem,
+                    // but then we'd have to save it in the database when scheduling. Can be done, not doing it now.
```
Top-Cat, line 37: Should probably not leave this comment in, but it is useful as a comment on the PR

An alternative (maybe easier, idk) approach could be to add a new state of `ScheduledWithAlert` or similar that triggers the alert but the "normal" `Scheduled` doesn't

Top-Cat (changes requested):

Top-Cat:
Ok to test

src/jvmMain/kotlin/io/beatmaps/util/scheduled.kt
Undeceiver: Copy/pasting the comment here then.

I am not adding the "alert" parameter to the database because it feels like too much effort for the unlikely case that someone scheduled a map for release but doesn't want to notify followers. If this ever happens, the alert will go out regardless (if it's scheduled it's likely it's a big deal). If we did want to consider the parameter, it can be passed here without any problem, but then we'd have to save it in the database when scheduling. Can be done, not doing it now.

Undeceiver: The reason I added it to the code was in case somebody stumbles upon this in the future, looks at that `true` and wonders why the hell is the alert flag not passed on. But I figure they can probably trace it back to the commit/PR if it happens.

PR #340, opened 2024-06-13
Collaboration option for followers

Undeceiver:
image

Currently the UI isn't working properly when unselecting, but this seems unrelated to this particular commit. I mentioned this to TC here:

https://discord.com/channels/882730837974609940/882731668589387796/1250611083719344180

### DIFF
--- a/src/commonMain/resources/db/migration/V70__FollowsCollabs.sql
+++ b/src/commonMain/resources/db/migration/V70__FollowsCollabs.sql
@@ -1,2 +1,4 @@
 ALTER TABLE public.follows
     ADD COLUMN "collab" boolean NOT NULL DEFAULT TRUE;
+
+UPDATE public.follows SET "collab" = "upload";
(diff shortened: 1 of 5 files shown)


src/commonMain/resources/db/migration/V70__FollowsCollabs.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE public.follows
+    ADD COLUMN "collab" boolean NOT NULL DEFAULT TRUE;
```
Top-Cat, line 2: Just for my sanity we should probably run an initial setup like
```sql
UPDATE public.follows SET "collab" = "upload";
```

So that people who are not following someone anymore aren't somehow set to collab. I know the query for alerts filters it out but there could be some weird side effects when changing follower preferences...

Top-Cat (changes requested):
I've also just rewritten the profile page (which didn't itself fix the checkbox bug, I've hacked around that with keys). I can help merge these changes with that if you want.

Undeceiver:
I'd appreciate it if you could help with the merge indeed. I'm not sure what the best way to do it with Git / GitHub is when there are conflicts. If I had to do it I'd probably look at the files that are conflicted and manually produce a combination, but I think that'd be annoying to get Git to agree on. Is there a systematic way to manually merge conflicts in GitHub?

Top-Cat (approved):

PR #346, opened 2024-06-24
Add map embeds

Pixelguymm:


src/jsMain/kotlin/io/beatmaps/maps/mapEmbed.kt
```diff
@@ -0,0 +1,49 @@
...
+        val mapKey = params["mapKey"]
+
+        setMap(null)
+
+        axiosGet<MapDetail>(
+            "${Config.apibase}/maps/id/$mapKey"
+        ).then {
+            setMap(it.data)
+        }.catch {
+            history.push("/")
```
Top-Cat, line 34: I really don't think you want to do this. Need some fancy error message instead.

The header isn't there and it's going to be weird having the homepage in an iframe...
Screenshot 2024-06-24 at 15 31 16

Top-Cat:
Should hide the add to playlist button in the embed as it requires the modal popup

Pixelguymm:
It didn't appear on foreign sites while testing, even when logged in. Did you get it to appear?

Top-Cat:
> It didn't appear on foreign sites while testing, even when logged in. Did you get it to appear?

Ah yes, I just went to `/embed` but of course the cookies won't work when actually in an iframe.

src/jsMain/kotlin/io/beatmaps/maps/mapEmbed.kt
Pixelguymm: Oh you're right, I'll see what I can do to fix that.

PR #353, opened 2024-06-28
Small playlist tweaks

Pixelguymm:


Top-Cat:
Before/after images or description pls 🙏

Pixelguymm:
Here you go. I added a little drag indicator when hovering over a map card, and made the delete button less huge.

**Before:**
chrome_Lscl4St61G

**After:**
chrome_yJt7EAEl02

PR #359, opened 2024-07-08
Hide links in dropdown when page is smol

Top-Cat:


src/jvmMain/sass/main.scss
```diff
@@ -1591,6 +1647,36 @@ small.alert-count {
       width: 44px;
     }
   }
+
+  .thin-dd {
+    .dropdown-divider {
+      display: none;
```
Pixelguymm, line 1653: I'd consider making these visible on larger screens as well, as vertical dividers that is. If you don't agree or want to leave that for another PR, everything else looks good to me.

Pixelguymm (approved):

src/jvmMain/sass/main.scss
Top-Cat: Can leave that for the future

PR #368, opened 2024-07-14
Fixes to recent alert features

Undeceiver:
Two fixes bundled in one:
- The "Notify followers" option will be **off by default** when updating a map that has previously been published, so as to avoid spamming users when a mapper updates a map several times in quick succession without unchecking the option. It remains on by default when it's the first time publishing the map, and in both cases the mapper can change it.
- I had not accounted for the fact that a collaboration could be accepted before a map was published. Now the "collaborated with" alert will only be sent if the collaboration is accepted **when the map is public**. Otherwise, whenever a map is published (including on updates), the alert will be sent to all followers of all mappers involved in the collaboration, following their individual follow preferences. I also added all mapper names to the collaboration text, so that users won't receive alerts for mappers they don't follow publishing maps (instead, it will show a list with all the mappers, which will include at least one that they follow).

I have extensively tested (and debugged) this. It actually took quite a bit of work to make this all work properly, but I'm quite confident now. Feel free to ask me about specific tests that you may think would be necessary and I can confirm to you whether I tried that.

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
@@ -130,7 +130,7 @@ fun Route.collaborationRoute() {
                                 .select(Follows.followerId)
                                 .where {
                                     followsAlias[Follows.id].isNull() and (Follows.followerId neq map.uploaderId) and
-                                            (Follows.userId eq sess.userId) and Follows.collab and Follows.following
+                                        (Follows.userId eq sess.userId) and Follows.collab and Follows.following
                                 }
                                 .map { row ->
                                     row[Follows.followerId].value
@@ -143,7 +143,7 @@ fun Route.collaborationRoute() {
                                         map.id.value
                                     )
                                 }: **${map.name}**.\n" +
-                                        "*\"${map.description.replace(Regex("\n+"), " ").take(100)}...\"*",
+                                    "*\"${map.description.replace(Regex("\n+"), " ").take(100)}...\"*",
                                 EAlertType.MapRelease,
                                 recipients
                             )
--- a/src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
@@ -115,32 +115,39 @@ fun Route.collaborationRoute() {
                             it[accepted] = true
                         }
 
-                        // Generate alert for followers of the collaborator.
-                        val followsAlias = Follows.alias("f2")
-                        val recipients = Follows
-                            .join(followsAlias, JoinType.LEFT, followsAlias[Follows.followerId], Follows.followerId) {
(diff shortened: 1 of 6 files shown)


Top-Cat:
> Can one of the admins verify this patch?

Add to whitelist

src/jvmMain/kotlin/io/beatmaps/util/versions.kt
```diff
@@ -125,16 +129,23 @@ fun publishVersion(mapId: Int, hash: String, alert: Boolean?, rb: RabbitMQInstan
...
 fun pushAlerts(map: BeatmapDao, rb: RabbitMQInstance?) {
+    val allAuthors = (map.collaborators.values + map.uploader).reversed()
+    val authorIds = allAuthors.map { it.id }
+
     val recipients = Follows.selectAll().where {
-        Follows.userId eq map.uploaderId and Follows.upload and Follows.following
+        (Follows.userId inList authorIds) and (Follows.followerId notInList authorIds) and (((Follows.userId eq map.uploaderId) and Follows.upload) or ((Follows.userId neq map.uploaderId) and Follows.collab)) and Follows.following
     }.map { row ->
         row[Follows.followerId].value
     }
```
Top-Cat, line 139: If a user follows more than one author this query returns them twice causing a duplicate key exception when inserting alerts below.

src/jvmMain/kotlin/io/beatmaps/util/versions.kt
```diff
@@ -125,16 +129,23 @@ fun publishVersion(mapId: Int, hash: String, alert: Boolean?, rb: RabbitMQInstan
...
 
+    val authorNames = allAuthors.map {
+        "@" + it.uniqueName
+    }.joinToString(separator = ", ")
```
Top-Cat, line 143: ⛳ This code should produce the same result

```kt
    val authorNames = allAuthors.joinToString {
        "@" + it.uniqueName
    }
```

src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
```diff
@@ -115,32 +115,39 @@ fun Route.collaborationRoute() {
...
-                                Integer.toHexString(
-                                    map.id.value
-                                )
-                            }: **${map.name}**.\n" +
-                                "*\"${map.description.replace(Regex("\n+"), " ").take(100)}...\"*",
-                            EAlertType.MapRelease,
-                            recipients
-                        )
+                        // Generate alert for followers of the collaborator, if the map has already been published.
+                        if (map.lastPublishedAt != null) {
```
Top-Cat, line 119: I'm not sure this is wrong but I'm interested in what you think _SHOULD_ happen when adding a collaborator to a previously published map that is currently _NOT_ published. ie it was unpublished.

Top-Cat (changes requested):
I wrote some test cases (which I've pushed to the dev branch) which currently just trigger an exception as described below. :(

PR #371, opened 2024-07-17
Publish reply events for the websocket

TimothyGillespie:
In combination with https://github.com/beatmaps-io/beatsaver-ws/pull/6

There may very well be a better way to do this. It was not immediately apparent to me, though.

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/api/review.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/review.kt
@@ -453,7 +453,7 @@ fun Route.reviewRoute() {
             val response = requireCaptcha(
                 reply.captcha,
                 {
-                    var insertedId: Int? = null;
+                    var insertedId: Int? = null
                     val response = newSuspendedTransaction {
                         val allowedUsers = Review
                             .join(Beatmap, JoinType.LEFT, Review.mapId, Beatmap.id)
@@ -483,7 +483,7 @@ fun Route.reviewRoute() {
                         ActionResponse(true, listOf())
                     }
 
-                    if(insertedId != null) {
+                    if (insertedId != null) {
                         call.pub("beatmaps", "ws.review-replies.created", null, insertedId!!)
                     }
 
@@ -502,7 +502,7 @@ fun Route.reviewRoute() {
             val update = call.receive<ReplyRequest>()
 
             captchaIfPresent(update.captcha) {
-                var updated = false;
+                var updated = false
                 val response = newSuspendedTransaction {
                     val ownerId = ReviewReply
                         .select(ReviewReply.userId)
@@ -543,7 +543,7 @@ fun Route.reviewRoute() {
                 }
 
                 // This should be outside the transaction - otherwise the websocket will send the old text
-                if(updated) {
+                if (updated) {
                     call.pub("beatmaps", "ws.review-replies.updated", null, req.replyId)
                 }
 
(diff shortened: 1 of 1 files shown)


src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -467,16 +468,26 @@ fun Route.reviewRoute() {
...
-                            ActionResponse(it.insertedCount > 0, listOf())
+                        try {
+                            insertedId = ReviewReply.insertAndGetId {
+                                it[userId] = user.userId
+                                it[reviewId] = req.reviewId
+                                it[text] = reply.text
+                                it[createdAt] = NowExpression(createdAt)
+                                it[updatedAt] = NowExpression(updatedAt)
+                            }.value
+                        } catch (e: Exception) {
```
Top-Cat, line 479: What's this try for? Probably shouldn't be catching everything...
If you're just trying to emulate the old code I wouldn't bother with it.

src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -453,7 +453,8 @@ fun Route.reviewRoute() {
             val response = requireCaptcha(
                 reply.captcha,
                 {
-                    newSuspendedTransaction {
+                    var insertedId: Int? = null
+                    val response = newSuspendedTransaction {
```
Top-Cat, line 457: val (insertedId, response) = newSuspendedTransaction {

src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -467,16 +468,26 @@ fun Route.reviewRoute() {
...
+                            }.value
+                        } catch (e: Exception) {
+                            return@newSuspendedTransaction ActionResponse(false, listOf())
                         }
+
+                        ActionResponse(true, listOf())
+                    }
+
+                    if (insertedId != null) {
+                        call.pub("beatmaps", "ws.review-replies.created", null, insertedId!!)
```
Top-Cat, line 487: Once you've made this a val you won't need the `!!` anymore

src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -491,6 +502,7 @@ fun Route.reviewRoute() {
             val update = call.receive<ReplyRequest>()
 
             captchaIfPresent(update.captcha) {
+                var updated = false
```
Top-Cat, line 505: Seems unnecessary when this is just `response.success`

Top-Cat (changes requested):

Top-Cat:
Also for jenkins... add to whitelist

src/jvmMain/kotlin/io/beatmaps/api/review.kt
TimothyGillespie: Yeah, you're right. I guess i can just fetch a SQLError (or what might apply here).
Previously, success was just determined by number of entries being greater. Because I need the I'd determine failure this way.

If you find that unnecessary I can remove it as well - I'm not sure how the API behaves, when an error is thrown. Just giving a 500 I'd guess?

src/jvmMain/kotlin/io/beatmaps/api/review.kt
TimothyGillespie: Definitely! This is so much smarter.

src/jvmMain/kotlin/io/beatmaps/api/review.kt
TimothyGillespie: I tried that initially, but overlooked something. Looking at it again I found my issue. Much better!

PR #373, opened 2024-07-17
Switch to server-side DESCing and client side sorting for API consistency

TimothyGillespie:
Sorry for the second PR.


TimothyGillespie:
nvm! Previous solution was the desired one. Just talking past each other.

PR #379, opened 2024-07-18
Allow collaborators to reply

### LINKED ISSUE #377: Allow Collaborators of Maps to Reply to Reviews as Well
Currently, only the uploader of the map and the author of the specific review can reply to a review.

It would be appropriate to allow collaborators to as their relation to the map (and thus the review) is often comparable to that of the uploader.

This task would be solved when **accepted** collaborators can reply to reviews.

TimothyGillespie:
Resolves #377


src/commonMain/resources/db/test/V72.2__AddCollaboratorsToMaps.sql
```diff
@@ -0,0 +1,17 @@
...
+    SELECT id INTO collab1Id FROM uploader WHERE "uniqueName" = 'collab1';
+    SELECT id INTO collab2Id FROM uploader WHERE "uniqueName" = 'collab2';
+
+    INSERT INTO collaboration ("mapId", "collaboratorId", "requestedAt", accepted, "uploadedAt") VALUES
+         (1, collab1Id, '2021-01-01 00:00:00', TRUE, '2021-01-01 00:00:00'),
+         (2, collab2Id, '2021-01-01 00:00:00', TRUE, '2021-01-01 00:00:00'),
+         (3, collab1Id, '2021-01-01 00:00:00', TRUE, '2021-01-01 00:00:00'),
+         (3, collab2Id, '2021-01-01 00:00:00', TRUE, '2021-01-01 00:00:00'),
+         (4, collab1Id, '2021-01-01 00:00:00', FALSE, '2021-01-01 00:00:00'),
+         (4, collab2Id, '2021-01-01 00:00:00', TRUE, '2021-01-01 00:00:00');
```
Top-Cat, line 16: Extra points for adding test users for sure _but_ the test maps aren't guaranteed to be maps 1-4 (although they probably will be) in staging they're actually 5-8 so this migration will probably fail :(

src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -491,14 +492,23 @@ fun Route.reviewRoute() {
...
+
+                        if (mapOfReview == null) {
+                            return@newSuspendedTransaction Pair(null, ActionResponse(false, listOf("Map not found")))
+                        }
+
+                        val allowedUsers: List<Int> = listOf(
+                            mapOfReview.uploaderId.value,
+                            *mapOfReview.collaborators.values.map { it.id.value }.toTypedArray(),
+                            review[Review.userId].value
+                        )
```
Top-Cat, line 511: I'm wondering why the choice of these two queries...

- Option A: Take the existing query and join onto collaborators.
Seems simpler and you're only bringing back 3 columns so the cross-product effect isn't too bad.
- Option B: Just add a query to the collaborators table. No cross product at all.

But this way you're bringing back the whole map `selectAll()` 😱 and then throwing away most of the data...

Top-Cat (changes requested):

src/commonMain/resources/db/test/V72.2__AddCollaboratorsToMaps.sql
TimothyGillespie: I see :thinking: I was wondering if I can assume that and at the same time it didn't really matter which one they are.

How about I just select the first 4 maps instead or ? :thinking:

src/jvmMain/kotlin/io/beatmaps/api/review.kt
TimothyGillespie: Yeah, this is not efficient. I was thinking the same - at the same time I don't expect this to be an actual issue performance wise.

Reason I did this was to use the complexToBeatmap function and have a list.

With Option A: If I just join collaborators, that might just be less efficient overall since joins are expensive and then I'd need to some odd, confusing parsing of the rows. So that didn't seem right.

Option B: Didn't consider that. So I'll probably do that :thinking:

TimothyGillespie:
I went ahead and implemented Option B and just use the first 4 maps now since it shouldn't matter too much which maps they are. Just in case, I made the INSERT INTO "ignore" conflicts.

src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -491,14 +492,32 @@ fun Route.reviewRoute() {
...
+                                ActionResponse(false, listOf("Review or map not found"))
+                            )
+                        }
+
+                        val collaborators = Collaboration
+                            .select(Collaboration.collaboratorId)
+                            .where { Collaboration.mapId eq mapId and Collaboration.accepted }
+                            .map { it[Collaboration.collaboratorId].value }
+
+                        val allowedUsers: List<Int> = listOf(
```
Top-Cat, line 516: Last nitpick, this explicit type isn't necessary

TimothyGillespie: I wasn't sure if it would surmount to the expected type in a previous iteration - so it served as a sanity check.
Has been removed now.

PR #380, opened 2024-07-18
Add notifications for replies

### LINKED ISSUE #378: Provide Notifications for Review Replies
Currently, replies to reviews will not add a notification to anyone resulting and are thus invisible to everyone involved unless they actively look for replies by going through all maps. That's not ideal.

This task would be done, when the following people receive notifications for replies:
- map uploader
- **accepted** map collaborator
- author of the review that has been replied too

TimothyGillespie:
Resolves #378 
Should be merged after #379 

Open Question: Should we add an alert for reviews for collaborators or remove the alert for collaborators when it comes to review replies?

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/api/review.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/review.kt
@@ -55,7 +55,6 @@ import org.jetbrains.exposed.sql.JoinType
 import org.jetbrains.exposed.sql.Query
 import org.jetbrains.exposed.sql.SortOrder
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.inList
-import org.jetbrains.exposed.sql.SqlExpressionBuilder.isNull
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.deleteWhere
 import org.jetbrains.exposed.sql.insertAndGetId
@@ -504,9 +503,12 @@ fun Route.reviewRoute() {
                             return@newSuspendedTransaction Pair(null, ActionResponse(false, listOf("Map not found")))
                         }
 
-                        val allowedUsers: List<Int> = listOf(
+                        val mapperIds = listOf(
                             mapOfReview.uploaderId.value,
-                            *mapOfReview.collaborators.values.map { it.id.value }.toTypedArray(),
+                            *mapOfReview.collaborators.values.map { it.id.value }.toTypedArray()
+                        )
+                        val allowedUsers: List<Int> = listOf(
+                            *mapperIds.toTypedArray(),
                             review[Review.userId].value
                         )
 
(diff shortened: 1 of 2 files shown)


src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -379,6 +380,19 @@ fun Route.reviewRoute() {
                             )
                             updateAlertCount(map.uploaderId.value)
                         }
+
+                        for (singleCollaborator in map.collaborators.values) {
```
Top-Cat, line 384: I'm not sure there's a single for loop in the rest of the codebase 😆

```kt
map.collaborators.values.filter { it.reviewAlerts }.forEach { singleCollaborator ->
```

src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -531,6 +551,37 @@ fun Route.reviewRoute() {
...
+                                        "*\"${reply.text.replace(Regex("\n+"), " ").take(100)}...\"*",
+                                    EAlertType.ReviewReply,
+                                    reviewUserId
+                                )
+
+                                updateAlertCount(reviewUserId)
+                            }
+
+                            for (singleCollaborator in collaborators) {
+                                if (singleCollaborator[collaboratorAlias[User.id]].value == user.userId) continue
```
Top-Cat, line 570: Does this work? User.id isn't in the select (although Collaboration.collaboratorId is)
I would have just kept the map above and made it
```kt
.map { it[Collaboration.collaboratorId].value to it[collaboratorAlias[User.reviewAlerts]] }
```
then here
```kt
filter { it.first != user.Id && it.second }
```

src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -492,30 +506,36 @@ fun Route.reviewRoute() {
...
 
-                        val allowedUsers = listOf(
+                        val mapperIds = listOf(
                             uploadUserId,
-                            *collaborators.toTypedArray(),
+                            *collaborators.map { it[Collaboration.collaboratorId].value }.toTypedArray()
+                        )
+
+                        val allowedUsers = listOf(
+                            *mapperIds.toTypedArray(),
```
Top-Cat, line 538: Is mapperIds used for anything other than composition back into allowedUsers 😕

Top-Cat (changes requested):

src/jvmMain/kotlin/io/beatmaps/api/review.kt
TimothyGillespie: In fact, it doesn't as I had come to notice too :man_facepalming: 
I change the part to a filter -> forEach too. I did not map it though.

Using `first` and `second` seemed a bit too unclear to me. I hope this works too, though.

src/jvmMain/kotlin/io/beatmaps/api/review.kt
TimothyGillespie: You're right! This PR became a bit more messy than intended.
It used to have another purpose - but it faded out over a few realizations and improvements.
Has been removed now.

TimothyGillespie:
This should be good now.

I've tested this one a bit more thoroughly again. It's possible this solves a few issues that have been introduced in #379, that I haven't caught in-between changes. Feeling pretty good about this one now.

src/jvmMain/kotlin/io/beatmaps/api/review.kt
```diff
@@ -379,6 +379,17 @@ fun Route.reviewRoute() {
...
                             updateAlertCount(map.uploaderId.value)
                         }
+
+                        map.collaborators.values.filter { it.reviewAlerts }.forEach { singleCollaborator ->
+                            Alert.insert(
+                                "New review on a map you collaborated on",
+                                "@${sess.uniqueName} just reviewed a map you collaborated on #${toHexString(updateMapId)}: **${map.name}**.\n" +
+                                    "*\"${newText.replace(Regex("\n+"), " ").take(100)}...\"*",
+                                EAlertType.Review,
+                                singleCollaborator.id.value
```
Top-Cat, line 389: I'm now wondering why this is a loop at all when you can pass a list here

Top-Cat: For extra context.

If you call `Alert.insert(` 5 times it will create 5 alerts and join that to 5 recipients
If you call `Alert.insert(` once with a list of 5 users it will only create the alert once and join it to all 5

Top-Cat: (updateAlertCount also takes a list)

TimothyGillespie: True. I was under the impression I need it for updateAlertCount. But I see it has a list based API as well!

Top-Cat (approved):

PR #388, opened 2024-07-27
Remove yourself from a collaboration.

Undeceiver:
image
image

### DIFF
--- a/src/jsMain/kotlin/io/beatmaps/maps/mapinfo.kt
+++ b/src/jsMain/kotlin/io/beatmaps/maps/mapinfo.kt
@@ -88,6 +88,7 @@ val mapInfo = fc<MapInfoProps> { props ->
     val userData = useContext(globalContext)
     val loggedInId = userData?.userId
     val isOwnerLocal = loggedInId == props.mapInfo.uploader.id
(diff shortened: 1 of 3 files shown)


src/jsMain/kotlin/io/beatmaps/maps/mapinfo.kt
```diff
@@ -88,6 +88,7 @@ val mapInfo = fc<MapInfoProps> { props ->
     val userData = useContext(globalContext)
     val loggedInId = userData?.userId
     val isOwnerLocal = loggedInId == props.mapInfo.uploader.id
+    val isCollaboratorLocal = if (props.mapInfo.collaborators != null) props.mapInfo.collaborators.let { loggedInId in it!!.map { collaborator -> collaborator.id } } else false
```
Top-Cat, line 91: It's normally best to avoid `!!`
Here kotlin can't smart cast inside the if because mapInfo is mutable and theoretically could change between time of check and time of use. (Probably not but hey...)

Generally I'd rewrite this as:
```kt
props.mapInfo.collaborators?.let { loggedInId in it.map { collaborator -> collaborator.id } } == true
```

Top-Cat:
Just the one comment to address :)

Ok to test

PR #400, opened 2024-08-12
Send websocket notification on collaborator accept and remove

TimothyGillespie:
On the request of @NSGolova. Seems to make a lot of sense to me.

### DIFF
--- a/src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
+++ b/src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
@@ -15,6 +15,7 @@ import io.beatmaps.common.dbo.UserDao
 import io.beatmaps.common.dbo.Versions
 import io.beatmaps.common.dbo.collaboratorAlias
 import io.beatmaps.common.dbo.joinUploader
+import io.beatmaps.common.pub
 import io.beatmaps.util.cdnPrefix
 import io.beatmaps.util.isUploader
 import io.beatmaps.util.requireAuthorization
@@ -119,6 +120,8 @@ fun Route.collaborationRoute() {
                             it[accepted] = true
                         }
 
+                        call.pub("beatmaps", "maps.${map.id}.updated.collaborators.added", null, map.id.value)
+
                         // Generate alert for followers of the collaborator, if the map has already been published.
                         if (published == true) {
                             val followsAlias = Follows.alias("f2")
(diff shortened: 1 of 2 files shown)


src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
```diff
@@ -119,6 +120,8 @@ fun Route.collaborationRoute() {
                             it[accepted] = true
                         }
 
+                        call.pub("beatmaps", "maps.${map.id}.updated.collaborators.added", null, map.id.value)
```
Top-Cat, line 123: Calling a publish from inside a transaction doesn't guarantee the transaction will be committed before the message is processed.

Top-Cat (changes requested):
What's the thinking behind having both `collaborators.added` and `collaborators.removed`? As they're not processed differently it's currently just adding complication.

src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
TimothyGillespie: You are right! I now pulled it out of the transaction - I hope they way I did is fine. Otherwise, happy to hear suggestions!

TimothyGillespie:
I did removed the `.added` and `.removed` respectively.

The reason why I did that is because it seemed "natural" to namespace it like that and if we were to refactor/change or utilize these somehow it'd be easier. I do find it to be clearer that way as to what is included etc.

But realistically, this will never get used and I see your point. Might have some negative performance or resource impact as well.

src/jvmMain/kotlin/io/beatmaps/api/collaboration.kt
```diff
@@ -152,17 +154,20 @@ fun Route.collaborationRoute() {
...
-                        false
+                        Pair(false, null)
                     }
                 } else {
-                    Collaboration.deleteWhere {
+                    val success = Collaboration.deleteWhere {
                         id eq req.collaborationId and (collaboratorId eq sess.userId)
                     } > 0
+
+                    Pair(success, null)
```
Top-Cat, line 166: But doesn't this mean that the message won't be sent? map == null
Given you can't get a map object here it might make more sense for the second result to also be a boolean and return `map != null` above instead of `map`

You'll also want to use `req.mapId` below instead of using the map object.

TimothyGillespie: Exactly! This is pretty much intended.

This is the case where the collaboration is rejected, hence no change. I find it bit awkward too though, best I could think of define the returned second parameter to be null if not successful accepted.

I did try to use `req.mapId`, but I quickly realized it doesn't exist, because the collaboration request object is denied or accepted and only that ID is passed.

PR #402, opened 2024-09-14
Change "name" to "uniqueName" in Sorting

TheCzar1994:
Should fix sort order issue for curators. Thanks @TimothyGillespie


Top-Cat:
Ok to test

PR #412, opened 2024-10-16
Add text-end to the map's infotable

TheCzar1994:
Extremely minor but this has always bugged me. Feel free to ignore/close this PR if you intended to have it the way it was or don't feel like making this change...
This should make the Mappers field align to the end of the row when the text wraps versus being the only field that goes to the left.
Before:
before
After:
after


Top-Cat:
Originally the text was not meant to wrap at all. I think this was changed when collaborators were added to the row.

Will double check this locally when I have time.

Top-Cat:
Ok to test

Top-Cat:
I'd go as far as to say maybe we could break up the field to better delineate mappers added as collaborators and the map info. But this is a good first step :)

PR #473, opened 2024-11-27
Environment filters

Pixelguymm:


src/jsMain/kotlin/io/beatmaps/index/main.kt
```diff
@@ -2,10 +2,12 @@ package io.beatmaps.index
 
 import io.beatmaps.Config.dateFormat
 import io.beatmaps.History
+import io.beatmaps.common.EnvironmentSet
 import io.beatmaps.common.MapTagSet
 import io.beatmaps.common.SearchOrder
 import io.beatmaps.common.SortOrderTarget
 import io.beatmaps.common.api.RankedFilter
+import io.beatmaps.common.toEnvironmentSet
```
Top-Cat, line 10: Where are this method and type 👀

Top-Cat: It's ok, I found them

PR #643, opened 2025-09-01
Add Map Check button to Map Cards

TheCzar1994:
Not sure if this is desired but... as suggested in site-suggestions in BeatSaver Discord. Only adds to main map pages, not on each individual card on the homepage. Uses the same icon Beat Leader is using to maintain consistency across sites. ~~Hopefully I did this all correctly.~~
Do we want the button somewhere else in the line? I prefer when pages open in new tabs so added the target = _blank but can remove if we don't like it.

Main map page:
image

Main map page on mobile:
image

Not visible on homepage cards:
image


TheCzar1994:
For some reason I didn't even think to just incorporate it by itself in the links... I like making extra work for myself. Will change it out here shortly.

PR #646, opened 2025-10-25
Adding line about AI generated songs

Undeceiver:
Note: I have not tested this. I don't currently have a local working copy of Beat Saver and felt it was unnecessary to set it up just for this. The commit involves one line in the ToS file and I did not believe it was necessary to test it. If you are not okay with this, reject the PR and let me know and I'll get the local copy installed.

This change follows the discussion here: https://discord.com/channels/882730837974609940/1407189628498350120


Top-Cat:
Ok to test

PR #683, opened 2026-06-29
Review silencing

Umbranoxio:
this adds a "silence" admin action, roughly inspired by osu's account standing / silence flow.

(for reference)
image

admins can now silence a user from reviewing maps for a set duration or permanently with a comment.

needs https://github.com/beatmaps-io/beatsaver-common-mp/pull/166

## what's in here

- new review silence moderation action
- admin UI for adding / managing silences
- review / reply checks so silenced users can't keep reviewing maps
- silence actions appear in the modlog
- profile account standing that combines relevant moderation history like silences & suspensions
- public "history pruning" for inactive actions via `PUBLIC_ACCOUNT_STANDING_DAYS` defaulting to 90 days

account standing is intentionally public by default for transparency **including** suspension / silence descriptions. lack of moderation transparency is something the community has been complaining about, so the public view is meant to make relevant recent / active account actions visible without exposing the full admin history forever. admins still see the full history and active / permanent actions "bypass" the public expiry since y'know, they're relevant.

BeatSaver may want to add durations to suspensions at some point too, but that's kinda out of scope for this PR and i'm already busy enough as it is with the world cup so i didn't touch suspensions

## screenshots
SCR-20260629-svjr
SCR-20260629-svxd
SCR-20260629-sxqx
SCR-20260629-taxj
SCR-20260629-tbga

---

this is my first time touching kotlin in years so be kind this was a very traumatic experience for me

### DIFF
--- a/modules/shared/src/commonMain/kotlin/io/beatmaps/api/users.kt
+++ b/modules/shared/src/commonMain/kotlin/io/beatmaps/api/users.kt
@@ -26,6 +26,18 @@ enum class PatreonTier(val pledge: Int, val supporting: Boolean, val title: Stri
     }
 }
 
+enum class AccountStandingAction {
+    SILENCE, SUSPENSION
+}
+
+@Serializable
+data class AccountStandingEntry(
+    val action: AccountStandingAction,
+    val createdAt: Instant,
+    val lengthMinutes: Int? = null,
+    val description: String? = null
+)
+
 @Serializable
 data class UserDetail(
     val id: Int,
@@ -49,7 +61,10 @@ data class UserDetail(
     val suspendedAt: Instant? = null,
     val playlistUrl: String? = null,
     val patreon: PatreonTier? = null,
-    val blurnsfw: Boolean? = null
+    val blurnsfw: Boolean? = null,
+    val reviewSilenced: Boolean = false,
+    val reviewSilencedUntil: Instant? = null,
+    val accountStanding: List<AccountStandingEntry> = emptyList()
 ) {
     fun profileLink(tab: String? = null, absolute: Boolean = false) = LinkHelper.profileLink(this, tab, absolute)
     companion object
@@ -117,13 +132,16 @@ data class AccountRequest(val currentPassword: String? = null, val password: Str
 @Serializable
 data class UserAdminRequest(val userId: Int, val maxUploadSize: Int, val maxVivifySize: Int, val curator: Boolean, val seniorCurator: Boolean, val curatorTab: Boolean, val verifiedMapper: Boolean) {
     companion object {
-        val allowedUploadSizes = arrayOf(0, 15, 30, 50)
-        val allowedVivifySizes = arrayOf(0, 5, 10, 20)
+        val allowedUploadSizes get() = arrayOf(0, 15, 30, 50)
+        val allowedVivifySizes get() = arrayOf(0, 5, 10, 20)
     }
 }
 
 @Serializable
 data class UserSuspendRequest(val userId: Int, val suspended: Boolean, val reason: String?)
+
+@Serializable
+data class UserReviewSilenceRequest(val userId: Int, val durationMinutes: Int?, val reason: String?)
 
 @Serializable
 data class UserFollowRequest(val userId: Int, val following: Boolean, val upload: Boolean, val curation: Boolean, val collab: Boolean)
--- a/modules/shared/src/commonMain/kotlin/io/beatmaps/api/users.kt
+++ b/modules/shared/src/commonMain/kotlin/io/beatmaps/api/users.kt
@@ -132,8 +132,8 @@ data class AccountRequest(val currentPassword: String? = null, val password: Str
 @Serializable
 data class UserAdminRequest(val userId: Int, val maxUploadSize: Int, val maxVivifySize: Int, val curator: Boolean, val seniorCurator: Boolean, val curatorTab: Boolean, val verifiedMapper: Boolean) {
     companion object {
-        val allowedUploadSizes get() = arrayOf(0, 15, 30, 50)
-        val allowedVivifySizes get() = arrayOf(0, 5, 10, 20)
+        val allowedUploadSizes = arrayOf(0, 15, 30, 50)
+        val allowedVivifySizes = arrayOf(0, 5, 10, 20)
     }
 }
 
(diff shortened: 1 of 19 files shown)


modules/shared/src/commonMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -117,14 +132,17 @@ data class AccountRequest(val currentPassword: String? = null, val password: Str
 @Serializable
 data class UserAdminRequest(val userId: Int, val maxUploadSize: Int, val maxVivifySize: Int, val curator: Boolean, val seniorCurator: Boolean, val curatorTab: Boolean, val verifiedMapper: Boolean) {
     companion object {
-        val allowedUploadSizes = arrayOf(0, 15, 30, 50)
-        val allowedVivifySizes = arrayOf(0, 5, 10, 20)
+        val allowedUploadSizes get() = arrayOf(0, 15, 30, 50)
+        val allowedVivifySizes get() = arrayOf(0, 5, 10, 20)
```
Top-Cat: This change appears unrelated to the purpose of the PR and as far as I can tell just makes this less efficient (new array on every get)

modules/shared/src/jsMain/kotlin/io/beatmaps/user/main.kt
```diff
@@ -527,104 +530,105 @@ val profilePage = fcmemo<Props>("profilePage") { _ ->
                     }
                 }
             }
+            val description = userDetail?.description?.take(500).orEmpty()
```
Top-Cat: I believe hiding the user's description when suspended was intentional. The other logic changes here mostly do nothing I think?

src/jvmMain/kotlin/io/beatmaps/api/reviewSilence.kt
```diff
@@ -0,0 +1,50 @@
...
+    val userId = reference("userId", User)
+    val moderatorId = reference("moderatorId", User)
+    val createdAt = timestamp("createdAt")
+    val silencedUntil = timestamp("silencedUntil").nullable()
+    val durationMinutes = integer("durationMinutes").nullable()
+    val reason = text("reason").nullable()
+    val revokedAt = timestamp("revokedAt").nullable()
+
+    private fun activeFor(userId: Int, now: java.time.Instant) =
+        (ReviewSilence.userId eq userId) and revokedAt.isNull() and (silencedUntil.isNull() or (silencedUntil greater now))
```
Top-Cat: Minor but I'd change this to take now as `now: Instant = Clock.System.now()` and call `now.toJavaInstant()` and then the calls below can simplify to `activeFor(userId)`

src/jvmMain/kotlin/io/beatmaps/api/reviewSilence.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.jetbrains.exposed.dao.id.IntIdTable
+import org.jetbrains.exposed.sql.SortOrder
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.greater
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.isNull
+import org.jetbrains.exposed.sql.and
+import org.jetbrains.exposed.sql.javatime.timestamp
+import org.jetbrains.exposed.sql.or
+
+object ReviewSilence : IntIdTable("review_silence", "silenceId") {
```
Top-Cat: By convention all tables live in the common repo.

src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -408,6 +420,66 @@ fun followData(uploaderId: Int, userId: Int?): UserFollowData {
...
 
+fun accountStanding(userId: Int, showAll: Boolean = false, suspended: Boolean = false): List<AccountStandingEntry> {
+    val now = Clock.System.now()
+    val nowInstant = now.toJavaInstant()
+    val publicCutoff = now.minus(publicAccountStandingDays).toJavaInstant()
+    fun visible(createdAt: java.time.Instant, active: Boolean) = showAll || active || createdAt >= publicCutoff
+
+    val silences = ReviewSilence
+        .selectAll()
+        .where { ReviewSilence.userId eq userId }
```
Top-Cat, line 446: This will likely table scan as you're not including `revokedAt == NULL` in the query.

src/jvmMain/kotlin/io/beatmaps/api/users.kt
```diff
@@ -408,6 +420,66 @@ fun followData(uploaderId: Int, userId: Int?): UserFollowData {
...
+        }
+
+    return (silences + suspensions).sortedByDescending { it.createdAt }.take(20)
+}
+
+fun UserDetail.withAccountStanding(userId: Int, showAll: Boolean = false) =
+    copy(
+        reviewSilenced = ReviewSilence.active(userId),
+        reviewSilencedUntil = ReviewSilence.activeUntil(userId)?.toKotlinInstant(),
+        accountStanding = accountStanding(userId, showAll, suspendedAt != null)
```
Top-Cat: It would be nice if as we're querying the table in accountStanding we could get all the data we need here in that single query?

Top-Cat (changes requested):
Firstly, thank you for making the effort to try this is overall pretty good and most of my code comments are minor.

I do wonder though if this timed system should entirely replace existing suspension. ie Remove the user suspended flag and have different types in this new table. Especially as using the ModLog table for the public standing is pretty dirty.

modules/shared/src/commonMain/kotlin/io/beatmaps/api/users.kt
Umbranoxio: My local instance exploded without this (probably something something to do in with my cursed local dev setup for this project)

Probably should've removed it as yeah, unrelated.

modules/shared/src/jsMain/kotlin/io/beatmaps/user/main.kt
Umbranoxio: oops, sorry, that change was leftover from an earlier iteration i did where account standing lived inside the profile info card and o was trying to work around the layout issues i caused in there, will revert

modules/shared/src/jsMain/kotlin/io/beatmaps/util/helpers.kt
```diff
@@ -18,6 +18,11 @@ fun <T> Promise<T>.orCatch(block: (Throwable) -> T) =
 fun <T> Promise<Promise<T>>.orCatch(block: (Throwable) -> T) =
     then({ it }, block)
 
+fun Throwable.responseErrors(): List<String> {
```
Umbranoxio, line 21: couldnt find a precedent anywhere in the codebase for this so here we are
if there is one and i missed it please point me to it

